PR #46, opened 2015-01-04
use imports.Process to correctly identify what imports are needed

ttacon:
Really gross just rereading file, processing and flushing it back out.


philhofer:
Thanks. Let's use this for now, and return to this issue if we can come up with an elegant way of keeping track of required imports.

PR #48, opened 2015-01-09
Type map

philhofer:
Get constant-time type information, including encoded size. 

This provides a small speedup in lots of places, and also reduces complexity and code size.

TODO: squash, etc. JSON translation will get a bit faster.

[Review on Reviewable](https://reviewable.io/reviews/philhofer/msgp/48)


philhofer:
Another easter egg: `(*Reader).ReadMapKeyPtr()` means that `DecodeMsg` methods no longer have to allocate at all. Bench time for those methods improves 10-50%.

PR #60, opened 2015-02-17
Arbitrary type support

philhofer:
Fixes #60 .

Attempting to always propagate type information into method generation has some nasty complexity drawbacks. Consider the following:

```go
type Insane [3]map[string]struct{A, B int}
``` 

I would argue that the generator *should* support the above, although, at its present state, it doesn't. 

This PR refactors everything to allow method generation on arbitrary types. Then, after processing each type individually, the generator can *choose* to inline the type's method.

Warning: this is still WIP.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/60)


philhofer:
TODO: make the following work correctly (right now `Chunk` is processed as IDENT)
```go
type Chunk [32]byte

// later...

type Thing struct {
    Data Chunk
}
```

PR #62, opened 2015-03-03
Number

philhofer:
Adds `msgp.Number` as a recognized concrete type that implements all of the standard interfaces. Similar to `json.Number`.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/62)


philhofer:
4f49ec0 also fixes a 50% perf hit for `EncodeMsg` methods on go1.5 tip caused by missing write barrier elision. I'll revert the change if the compiler starts eliding re-slicing write barriers.

ttacon:
Might as well make msgp.Number implement Stringer (json.Number does)?

PR #63, opened 2015-03-25
support reading int from uint lead byte

thinxer:
Please refer to https://github.com/msgpack/msgpack-c/issues/247 and https://github.com/msgpack/msgpack/issues/164.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/63)

### DIFF
--- a/msgp/read.go
+++ b/msgp/read.go
@@ -556,6 +556,42 @@ func (m *Reader) ReadInt64() (i int64, err error) {
 		i = getMint64(p)
 		return
 
+	case muint8:
+		p, err = m.r.Next(2)
+		if err != nil {
+			return
+		}
+		i = int64(getMuint8(p))
+		return
+
+	case muint16:
+		p, err = m.r.Next(3)
+		if err != nil {
+			return
+		}
+		i = int64(getMuint16(p))
+		return
+
+	case muint32:
+		p, err = m.r.Next(5)
+		if err != nil {
+			return
+		}
+		i = int64(getMuint32(p))
+		return
+
+	case muint64:
+		p, err = m.r.Next(9)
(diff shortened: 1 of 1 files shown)


philhofer:
There's an explanation in the Wiki for why I have [forbidden this sort of transformation thusfar](https://github.com/tinylib/msgp/wiki/Type-Mapping-Rules). In short, neither `int` cannot be safely decoded into `uint`, and `uint` cannot be safely decoded into `int.` Additionally, it is trivial for the user to use `(*Decoder).NextType()` to see whether or not the next object is encoded as a `uint`.

I understand that some other implementations always encode positive integers as `uint`s, but I believe that is a mistake on their part. (My opinion is that languages that do not support unsigned integers shouldn't be using `uint` encoding *at all*.)

Let's ping @ttacon and see what he thinks.

If this is going to change, though, then you'll want to make a similar change to `ReadInt64Bytes` in `read_bytes.go`.

ttacon:
IMHO, there isn't really a need for this "convenience" functionality. My mind could be changed, but a convincing argument would have to be made. I agree with @philhofer on this, I don't really think it's a good idea.

philhofer:
I would also add that `msgp.Number` can read any `int`, `uint`, `float32`, or `float64` for precisely that reason. The code generator understands that type natively.

philhofer:
I have added [a stub in the Wiki to explain how this should be handled.](https://github.com/tinylib/msgp/wiki/Tips-and-Tricks)

thinxer:
The discussion https://github.com/msgpack/msgpack/issues/164 states that format codes should not be used as type qualifiers. We have to fix this.

PR #64, opened 2015-03-25
move circular to testing

thinxer:
This prevents testing flags from appearing in normal builds.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/64)


thinxer:
Oops. Have to think more about where to put EndlessReader. In general it is only for testing and should not appear in normal builds.

Issue #34, opened 2014-11-23
Ability to specify which methods are generated on a per struct basis


myitcv:
e.g. in pseudo code:

```go
//msgp:shim string as:[]byte using:[]byte/string

//msgp:tuple MyTuple generate:DecodeMsg

type MyTuple struct {
    A, B string
}
```

would cause only the `DecodeMsg` method to be generated (but not any of the marshall-related methods or the `EncodeMsg`)

The use case here is again MSGPACK RPC. I want to be able to generate a decoder for one `struct` (the args to the method) and an encoder for another (the return values). Then combine those two structs into a wrapper `struct` along the following lines:

```go
//msgp:tuple MethodArgs generate:DecodeMsg

type MethodArgs struct {
  I int
}

//msgp:tuple MethodRetVals generate:EncodeMsg

type MethodRetVals struct {
  S string
}

type MethodWrapper struct {
  *MethodArgs
  *MethodRetVals
}
```

I can then unambiguously call `DecodeMsg` on an instance of the `MethodWrapper`

Thoughts?

PR #72, opened 2015-04-03
Made it configurable on a per file basis if the encode funcs have pointer or value receivers

### LINKED ISSUE #68: Use build tags or similar to enable using msgp without using unsafe package
App Engine doesn't allow importing unsafe, so right now it's impossible to use this lib there :/

### LINKED ISSUE #71: Make the encode funcs bound to the types, not pointers the types.
Because most encoding frameworks allow encoding either values or pointers to values (even if they only allow decoding to pointers for obvious reasons), it would ease integration with msgp if the encoder funcs were bound to values instead of pointers.

As in:

func (t Type) MarshalMsg ...

vs

func (t *Type) MarshalMsg ...

Maybe you have good reasons for having pointer receivers, but since I guess most of any type gets accessed in the encode func anyway, the copy shouldn't (without me having benchmarked this at all :) be slower than doing lots of dereferencing in a pointer receiver func.

Also, the value receiver func will work even when you have a *t, but not the reverse, so it's easier on the user :D 

Does this make sense, or am I missing something?

zond:
Resolves #71 in the simplest possible way.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/72)

### DIFF
--- a/gen/decode.go
+++ b/gen/decode.go
@@ -36,7 +36,7 @@ func (d *decodeGen) Execute(p Elem) error {
 
 	d.p.comment("DecodeMsg implements msgp.Decodable")
 
-	d.p.printf("\nfunc (%s %s) DecodeMsg(dc *msgp.Reader) (err error) {", p.Varname(), methodReceiver(p))
+	d.p.printf("\nfunc (%s %s) DecodeMsg(dc *msgp.Reader) (err error) {", p.Varname(), methodReceiver(p, false))
 	next(d, p)
 	d.p.nakedReturn()
 	unsetReceiver(p)
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -129,12 +129,17 @@ var builtins = map[string]struct{}{
 }
 
 // common data/methods for every Elem
-type common struct{ vname, alias string }
+type common struct {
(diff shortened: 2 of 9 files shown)


philhofer:
I think we're tickling a larger structural issue here, which is that shims cannot apply directly to one generation pass. (Consider how opaque things could get if we get more feature requests for method-specific options!) There's still an outstanding issue ( #34 ) for being able to turn a particular method pass off on a given type, so an ideal solution would solve both of these problems.

I'll open an issue for method-specific directives.

zond:
Would it be possible to allow this pull request onboard until you find a way to tackle the structural problems? It would be nice to be able to use main line msgp instead of my own fork :)

zond:
Never mind, I just made it all work with my branch, and my first trivial timing of the new performance showed worse performance.

Obviously this is a problem somewhere in my code, since I find it unlikely that your generated encoders/decoders are slower than the reflective encoder/decoder of https://github.com/vmihailenco/msgpack, but it is sufficiently annoying that I will give this up for a few days at least.

Damnit...

philhofer:
I'm going to close this for now, but ping me if you want it re-opened.

PR #76, opened 2015-04-06
handle 0xff (negative fixed int -1) when setting up sizes

cce:
In the `init()` for elsize.go there is a loop that sets up sizes for recognized negative fixed int types, but it does not add an entry for `0xff` (negative fixed integer -1) because the loop is written as 
`for i := mnfixint; i < 0xff; i++`.  

This fixes that bug — without it msgp will produce "msgp: unrecognized type prefix 0xff" (InvalidPrefixError) when trying to read msgpack data with negative fixed integer value -1.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/76)

### DIFF
--- a/msgp/elsize.go
+++ b/msgp/elsize.go
@@ -48,8 +48,8 @@ func init() {
 	}
 
 	// nfixint
-	for i := mnfixint; i < 0xff; i++ {
-		sizes[i] = bytespec{size: 1, extra: constsize, typ: IntType}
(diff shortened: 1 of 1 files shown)


msgp/elsize.go
```diff
@@ -48,8 +48,8 @@ func init() {
 	}
 
 	// nfixint
-	for i := mnfixint; i < 0xff; i++ {
```
philhofer: Would you mind just changing this to `i <= 0xff`?

PR #87, opened 2015-06-03
Always generate msgp.Sizer if -marshal is true.

### LINKED ISSUE #86: Generated code misses Msgsize() implementations when running with -io=false
$ cat x.go
    package x

    //go:generate msgp -io=false

    type X int

    $ go generate
    ...
    >>> Formatted "x_gen.go"
    >>> Done.
    $ go build
    # _/tmp
    ./x_gen.go:13: z.Msgsize undefined (type X has no field or method Msgsize)

ebfe:
MarshalMsg() depends on Msgsize() but that is only generated when `-io=true`.

Fixes #86

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/87)

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,7 +60,7 @@ func main() {
 		mode |= (gen.Encode | gen.Decode | gen.Size)
 	}
 	if *marshal {
-		mode |= (gen.Marshal | gen.Unmarshal)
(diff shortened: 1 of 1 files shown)


philhofer:
Nice. Thanks.

PR #89, opened 2015-06-06
Format generated tests

ebfe:
Run the test code through imports.Process as it is done with the regular code.
This makes #88 obsolete as the unused import is fixed automatically.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/89)


philhofer:
While I was reading this, I had an idea to make things a little more efficient from a syscall standpoint, so I included this change in #90. Thanks for the PR, though; I really appreciate your contributions to the project!

PR #95, opened 2015-06-15
add option to allow processing of unexported types

ebfe:
[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/95)

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,11 +36,12 @@ import (
 )
 
 var (
-	out     = flag.String("o", "", "output file")
-	file    = flag.String("file", "", "input file")
-	encode  = flag.Bool("io", true, "create Encode and Decode methods")
-	marshal = flag.Bool("marshal", true, "create Marshal and Unmarshal methods")
-	tests   = flag.Bool("tests", true, "create tests and benchmarks")
+	out        = flag.String("o", "", "output file")
+	file       = flag.String("file", "", "input file")
+	encode     = flag.Bool("io", true, "create Encode and Decode methods")
+	marshal    = flag.Bool("marshal", true, "create Marshal and Unmarshal methods")
+	tests      = flag.Bool("tests", true, "create tests and benchmarks")
+	unexported = flag.Bool("unexported", false, "also process unexported types")
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


ebfe:
Consider this an RFC. Things are named pretty awkwardly. Suggestions welcome.

main.go
```diff
@@ -41,6 +41,7 @@ var (
 	encode  = flag.Bool("io", true, "create Encode and Decode methods")
 	marshal = flag.Bool("marshal", true, "create Marshal and Unmarshal methods")
 	tests   = flag.Bool("tests", true, "create tests and benchmarks")
+	exports = flag.Bool("exports", true, "process only exported types")
```
ttacon: Personally, I'd rather this be the other way. Otherwise usage is a little awkward. For example:

```
msgp -exports=false ...
```

is far less useable/readable as opposed to if you defined the flag to be something along the lines:

```go
var allTypes = flag.Bool("all-types", false, "process all types")
```

which means usage would be

```
msgp -all-types ...
```

which is far more readable IMHO, but that's just my two cents.

parse/getast.go
```diff
@@ -23,10 +24,11 @@ type FileSet struct {
 
 // File parses a file at the relative path
 // provided and produces a new *FileSet.
-// (No exported structs is considered an error.)
+// (No structs is considered an error.)
```
ttacon: Doesn't this maintain the previous behavior if the flag to process local imports isn't provided? The documentation should make this clear, currently it makes me think that this would change the previous behaviour.

main.go
philhofer: I agree. `unexported` also works.

parse/getast.go
philhofer: It shouldn't say "structs," either, since we've been able to process most composite types for a while now.

philhofer:
Sorry it took me a while to find time to look at this.

If `unexported` (or whatever) is a global processing flag, then it doesn't need to be in `modes.` (Those are for generation paths that can be turned on/off per type declaration.)

ebfe:
Updated. I think I addressed most (all?) of the comments.

Unfortunately another problem showed up. The generated benchmarks/tests for
unexported types are not picked up by `go test`.

ebfe:
I think the problem with benchmarks/tests is that the function names now begin with Test[lowercase] while `go test` only looks for Test[uppercase] prefixes.

philhofer:
I think the solution is to invert the naming of the existing tests. Instead of `Test[name]Unmarshal`, we should generate `TestUnmarshal[name]`.

ebfe:
Done

philhofer:
Great. LGTM.

PR #106, opened 2015-07-20
Break dependency on "testing" package

alexandercampbell:
Importing the "testing" standard library package automatically registers the `testing` [commandline flags](http://golang.org/src/testing/testing.go#L170). This side effect cluttered up the output of --help for any program that imported msgp.

I replaced the dependency on testing.B with an interface containing the required methods.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/106)

### DIFF
--- a/msgp/circular.go
+++ b/msgp/circular.go
@@ -1,20 +1,21 @@
 package msgp
 
-import (
-	"testing"
-)
+type timer interface {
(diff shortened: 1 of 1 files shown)


philhofer:
Good catch. Thanks!

PR #108, opened 2015-07-21
De-randomize type order in generated code

### LINKED ISSUE #107: consider generating structs/methods in a canonical order
I'm loving tinylib/msgp. Thanks so much for this fabulous library.

A minor nit: running go generate msgp seems to write the structs and methods within the generated code files in a random order, which causes git to think that they've been updated even when they really haven't.

Not a big deal. Can't think of an easy fix off the top of my head. Some kind of internal alphabetical sort before writing perhaps? Or perhaps lexical ordering following that of the source file?

Jason

ebfe:
Fixes #107

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/108)

### DIFF
--- a/parse/getast.go
+++ b/parse/getast.go
@@ -2,14 +2,16 @@ package parse
 
 import (
 	"fmt"
-	"github.com/tinylib/msgp/gen"
-	"github.com/ttacon/chalk"
 	"go/ast"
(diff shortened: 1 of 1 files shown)


philhofer:
Perfect.

Issue #109, opened 2015-07-21
how to resume I/O truncated message decoding?


glycerine:
2e776fce9d47cb00819f6dc939a708dbb3085cb6

Am I using DecodeMsg() wrong or is this a bug?  It is highly likely I may not be using DecodeMsg() properly; how should one recover from failed attempt to read a truncated message?

I'm trying to recover after a torn read, but I'm getting `msgp: attempted to decode type "int" with method for "bin"`. The test below fails when truncated after the 2nd byte.

~~~
package main

import (
	"bytes"
	"fmt"

	"github.com/tinylib/msgp/msgp"
)

//go:generate msgp

type ByteSlice []byte

func panicOn(err error) {
	if err != nil {
		panic(err)
	}
}

func chkb(a, b []byte) {
	if len(a) != len(b) {
		panic("chkb: len differ")
	}
	for i := range a {
		if a[i] != b[i] {
			panic("chkb: bytes differ")
		}
	}
}

var p = fmt.Printf

func main() {
	// This is a test of torn messages: truncated due to I/O pause, and then later resumed.
	//
	// Here we do a simple msgpack encoding of a 3 byte payload in []byte, called a ByteSlice
	// to get the msgp generated methods.
	//
	// The following is a test of decoding bin8 when the outmost []byte message
	// has been truncated.
	//
	// Given that we may read a ByteSlice ([]byte) msgpack message
	// that can be torn (io stall) at any byte boundary, when try to
	// decode any number of bytes of a message that is torn, we should get
	// an error, even if this happens multiple times.
	//
	// More: When the full bytes do become available we should be able to
	// pick up and read the whole message sucessfully; we should proceed where we left off.

	payload := ByteSlice([]byte{0x1, 0x2, 0x3})
	msgpacked, err := payload.MarshalMsg(nil)
	panicOn(err)

	n := len(msgpacked)
	p("\n len of msgpacked,  n = %d.  message = '%#v' should be bin8/0xc4\n", n, msgpacked)
	//len of msgpacked,  n = 5.  message = '[]byte{0xc4, 0x3, 0x1, 0x2, 0x3}' should be bin8/0xc4

	buf := bytes.NewBuffer(msgpacked)
	mpr := msgp.NewReader(buf)

	var payload2 ByteSlice
	err = payload2.DecodeMsg(mpr)
	panicOn(err)

	chkb([]byte(payload), []byte(payload2))
	p("\nuntorn payload message comes back fine.\n")

[...]

Issue #91, opened 2015-06-09
Interface namespace collision


philhofer:
`(*Writer).WriteString` and `(*Writer).WriteByte` collide with `io.ByteWriter` and `io.StringWriter`.

philhofer:
For `msgp.Reader`, this is solved by #110

PR #110, opened 2015-07-21
msgp.Reader.r -> msgp.Reader.R. Allow client m.R.Peek(). workaround #109

glycerine:
Makes the msgp.Reader member r *fwd.Reader into R, making it
public instead of private so clients of msgp can call m.R.Peek
and handle torn objects/EOF on streams gracefully. Allows
clients of msgp to work around #109 and is in general useful
to be able to Peek ahead in the data stream, leveraging the
useful properties of the fwd library from the application.

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/110)

### DIFF
--- a/msgp/extension.go
+++ b/msgp/extension.go
@@ -226,7 +226,7 @@ func (mw *Writer) WriteExtension(e Extension) error {
 // peek at the extension type, assuming the next
 // kind to be read is Extension
 func (m *Reader) peekExtensionType() (int8, error) {
-	p, err := m.r.Peek(2)
+	p, err := m.R.Peek(2)
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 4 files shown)


philhofer:
Hey -- thanks for the PR.

Exporting the field seems like a big API promise to make. I'd rather expose the `Peek()`/`Next()`/etc. methods through thunks. (Also, philosophically, I've generally been opposed to exposing non-stdlib dependencies to users directly, because you've essentially locked yourself into that dependency for the rest of a major version cycle, and it's another library that the user is implicitly supposed to grok.)

philhofer:
I've thought about it, and I changed my mind. This actually creates a work-around for #91 (e.g. (*Reader).ReadByte()` vs. `(*fwd.Reader).ReadByte()`)

PR #111, opened 2015-07-22
Add methods to read/write str/bin headers

philhofer:
In conversation with #109 .

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/111)

### DIFF
--- a/msgp/read.go
+++ b/msgp/read.go
@@ -759,6 +759,45 @@ func (m *Reader) ReadBytes(scratch []byte) (b []byte, err error) {
 	return
 }
 
+// ReadBytesHeader reads the size header
+// of a MessagePack 'bin' object. The user
+// is responsible for dealing with the next
+// 'sz' bytes from the reader in an application-specific
+// way.
+func (m *Reader) ReadBytesHeader() (sz uint32, err error) {
+	var p []byte
+	p, err = m.r.Peek(1)
+	if err != nil {
+		return
+	}
+	switch p[0] {
+	case mbin8:
+		p, err = m.r.Next(2)
+		if err != nil {
+			return
+		}
+		sz = uint32(p[1])
+		return
+	case mbin16:
+		p, err = m.r.Next(3)
+		if err != nil {
+			return
+		}
+		sz = uint32(big.Uint16(p[1:]))
+		return
+	case mbin32:
+		p, err = m.r.Next(5)
+		if err != nil {
+			return
+		}
+		sz = uint32(big.Uint32(p[1:]))
+		return
+	default:
+		err = badPrefix(BinType, p[0])
+		return
+	}
+}
+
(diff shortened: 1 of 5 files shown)


msgp/read.go
```diff
@@ -759,6 +759,45 @@ func (m *Reader) ReadBytes(scratch []byte) (b []byte, err error) {
...
+// way.
+func (m *Reader) ReadBytesHeader() (sz uint32, err error) {
+	var p []byte
+	p, err = m.r.Peek(1)
+	if err != nil {
+		return
+	}
+	switch p[0] {
+	case mbin8:
+		p, err = m.r.Next(2)
```
glycerine: On its face it appears these Next() calls would preclude restarting the read of the header on failure.

philhofer: [`Next()` does not alter the parse state if an error is returned.](https://github.com/philhofer/fwd/blob/master/reader.go#L221)

But if you read the header successfully, then yes, you can't restart back before the header.

glycerine: I find it cleaner to treat the reading of the header and the reading of the bytes that follow as one unit in terms of recovery from EOF. The variable length header means I've got to reparse those bytes anyway. If either fails I'd prefer to just start all over.

philhofer: If you want to recover from some point in history, you need to keep all of the bytes between that point and the present around in a buffer somewhere. That opens up the possibility for some catastrophic worst case performance, so the "right" solution here is going to be application-dependent. (For example: I've got an application that streams `bin` objects with sizes on the order of tens of MBs. Re-starting a stream 10MB in the past would be catastrophically expensive.)

I'm anticipating that people will use this feature something like the following:

```go
func isFatal(err error) bool {  return err != io.EOF } 

func ReadLargeFlakyBin(sink io.Writer, src *msgp.Reader) error {
   var sz uint32
   var err error
   for {
      sz, err = src.ReadBinHeader()
      if err != nil {
          if isFatal(err) {
              return err
          }
          continue
      }
      break
   }
   tot := int64(sz)
   var n int64
   for tot > 0 {
       n, err = io.CopyN(sink, src, tot)
       tot -= n
       if err != nil && isFatal(err) {
           return err
       }
   }
   return nil
}
```
In the example above, the total memory use is bounded by whatever `sink` buffers in memory.

glycerine: hi @philhofer--that makes sense. My messages are 100 bytes +/- 20 bytes. After an io.Copy the last one of them read in will have been truncated, 99% of the time. For my app its a common need to recover a bin that is ~100 bytes easily and without fuss.

I understand if you don't want to support two different transactional read modes.  In that case I'd appreciate it if you'd merge PR #110 so I don't have to maintain a separate branch to move forward, but even that isn't really a big deal. It's already branched and I've moved on to the next thing, so either way is really fine.

PR #130, opened 2015-08-24
Fix variable redeclaration in nested tuples

### LINKED ISSUE #129: Nested msgp:tuple leads to redeclared variable
For example:

```go
//msgp:tuple GetTwoNumbersArgs
type GetTwoNumbersArgs struct {
	FunctionArgs GetTwoNumbersFunctionArgs
}

//msgp:tuple GetTwoNumbersFunctionArgs
type GetTwoNumbersFunctionArgs struct {
	Arg0 int64
}
```

results in the generated code:

```go
func (z *GetTwoNumbersArgs) DecodeMsg(dc *msgp.Reader) (err error) {
	var ssz uint32
	ssz, err = dc.ReadArrayHeader()
	if err != nil {
		return
	}
	if ssz != 1 {
		err = msgp.ArrayError{Wanted: 1, Got: ssz}
		return
	}
	var ssz uint32
	ssz, err = dc.ReadArrayHeader()
	if err != nil {
		return
	}
	if ssz != 1 {
		err = msgp.ArrayError{Wanted: 1, Got: ssz}
		return
	}
	z.FunctionArgs.Arg0, err = dc.ReadInt64()
	if err != nil {
		return
	}
	return
}
```

which fails to compile because of the redeclared `ssz`

PS `msgp:tuple` is working like a dream otherwise!

ebfe:
Fixes #129

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/130)

### DIFF
--- a/gen/decode.go
+++ b/gen/decode.go
@@ -82,9 +82,11 @@ func (d *decodeGen) assignAndCheck(name string, typ string) {
 func (d *decodeGen) structAsTuple(s *Struct) {
 	nfields := len(s.Fields)
 
+	d.p.print("\n{")
 	d.p.declare(structArraySizeVar, u32)
 	d.assignAndCheck(structArraySizeVar, arrayHeader)
 	d.p.arrayCheck(strconv.Itoa(nfields), structArraySizeVar)
+	d.p.print("\n}")
 	for i := range s.Fields {
 		if !d.p.ok() {
 			return


myitcv:
Thanks @ebfe works for me locally.

gen/decode.go
```diff
@@ -85,12 +85,14 @@ func (d *decodeGen) structAsTuple(s *Struct) {
 	d.p.declare(structArraySizeVar, u32)
 	d.assignAndCheck(structArraySizeVar, arrayHeader)
 	d.p.arrayCheck(strconv.Itoa(nfields), structArraySizeVar)
+	d.p.print("\n{")
```
philhofer: I think the generated code would look a bit cleaner if we put `declare(...); assignAndCheck(...); arrayCheck(...)` inside a block, rather than putting each field in a block. It may look a little strange in the degenerate case (deeply-nested tuples), but in the common case we won't be needlessly declaring many new blocks.

philhofer:
Thanks for this fix! Just one in-line comment about which code gets its own block.

gen/decode.go
ebfe: > I think the generated code would look a bit cleaner if we put
> `declare(...); assignAndCheck(...); arrayCheck(...)` inside a block,
> rather than putting each field in a block. 

True, updated.

Issue #132, opened 2015-09-10
Not support pointer *int, *bool, etc.


mengzhuo:

glycerine:
it rarely makes sense to serialize pointers, as the next process that reads them will have a different address space layout.

what are you trying to accomplish?

mengzhuo:
I'm trying to marshal an rpc.response interface which in the standard library.

philhofer:
You're right that the `WriteIntf`/`ReadIntf` family of functions don't deal with pointers-to-ints, etc. (The documentation says as much.) I'm not sure they should be extended to support pointers-to-whatever. (In retrospect, I'm not sure whether or not I should have written those functions at all, since they support a very limited set of types, and people consistently stumble over their use as a consequence.)

There is a work-around, though: you can write a function that does a type switch over all the types you expect to support and serialize the correct concrete type. The library gives you all the tools you need to do this.

The other possibility here is to replace `[]interface{}` with `[]msgp.Marshaler` (or similar) in your code, which guarantees statically that you have something serializable. That's what I've done in [my own rpc code](http://github.com/tinylib/synapse).

mengzhuo:
@philhofer  You are right that we shouldn't support "pointers to whatever"
However I think this is a bug.
https://github.com/tinylib/msgp/blob/master/msgp/write.go#L612
"A pointer to a supported type"
*int which is pointer to int, which we should supported.

philhofer:
Fixed by #133

PR #133, opened 2015-09-12
fix AppendIntf missing pointer direction #132

mengzhuo:
```go
package t
//go:generate msgp -o t.go

import "testing"

type I struct {
        I interface{}
}

func TestInterfaceInt(t *testing.T) {

        b := 5
        i := &I{&b}
        _, err := i.MarshalMsg(nil)
        if err != nil {
                t.Error(err)
        }
}
```


philhofer:
Perfect. Thank you.

PR #135, opened 2015-10-23
Add some explicit type declarations to fix 32bit

tianon:
Happy to rebase/amend/modify/adjust as desired. :+1:

This allows 32bit compilation/testing to succeed, which otherwise fails like this:
```console
root@6a350063d957:/go/src/github.com/tinylib/msgp# go test ./msgp 
# github.com/tinylib/msgp/msgp
msgp/write_bytes_test.go:17: constant -9223372036854775808 overflows int
msgp/write_bytes_test.go:30: constant -9223372036854775808 overflows int
msgp/write_test.go:15: constant 2147483747 overflows int
FAIL	github.com/tinylib/msgp/msgp [build failed]
```

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/135)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: go
 
 go:
   - 1.4
+  - 1.5
   - tip
+
+env:
+  - GIMME_ARCH=amd64
+  - GIMME_ARCH=386
 
 script: "make travis"
--- a/Makefile
+++ b/Makefile
@@ -46,4 +46,10 @@ get-deps:
 all: install $(GGEN) $(MGEN)
 
 # travis CI enters here
-travis: get-deps test
+travis:
+	go get -d -t ./...
+	go build -o "$${GOPATH%%:*}/bin/msgp" .
+	go generate ./msgp
+	go generate ./_generated
+	go test ./msgp
+	go test ./_generated
--- a/msgp/write_bytes_test.go
+++ b/msgp/write_bytes_test.go
@@ -14,7 +14,7 @@ func TestIssue116(t *testing.T) {
 		t.Fatal(err)
 	}
 	if i != math.MinInt64 {
-		t.Errorf("put %d in and got %d out", math.MinInt64, i)
+		t.Errorf("put %d in and got %d out", int64(math.MinInt64), i)
 	}
 
 	var buf bytes.Buffer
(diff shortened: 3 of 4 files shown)


tianon:
Hitting a little bit of a chicken-and-egg issue with the `386` builds in Travis that I can't quite figure out what the best way to resolve is (where we need to `go generate` before we can build the `msgp` binary, but we need the `msgp` binary to `go generate`).  I'm not actually sure why the `amd64` builds succeed -- do the Travis workers somehow already have the `msgp` binary available but only on `amd64`? :confused:

philhofer:
Yeah, that looks like an oversight in the Makefile. The `go get ./...` does an implicit install for the host OS, but apparently not for cross-compilation.  (All the generated code is test code, so it shouldn't be a build dependency for the tool, but I goofed.)

Can you try changing the `travis` rule to:
```
travis:
    go get -d -t ./...
    go install .
    go generate ./msgp
    go generate ./_generated
    go test ./msgp
    go test ./_generated
```

Thanks!

tianon:
Sure, I'll go make that change :smile:

As an aside, `go get -d` is _supposed_ to be instructing `go` not to install:
```
The -d flag instructs get to stop after downloading the packages; that is,
it instructs get not to install the packages.
```

So I'm even more confused.

philhofer:
Bizarre.

Looks like `go install` doesn't put the binary in the expected place when compiling for 386. May have to switch to invoking `go build -o`... (Ugh. Sorry.)

tianon:
Yeah, no worries; I don't mind iterating! :smile:

tianon:
Success! :tada:

Still happy to rebase/amend/etc as desired. :+1:

philhofer:
Awesome, thanks. Could you squash 30cd4b6 and 421c096 together?

tianon:
Definitely, done! :metal:

tianon:
Thanks :heart:

tianon:
As a total side note, considering a 1.0-beta or maybe 1.0 release soon? :smile:

philhofer:
[Yep!](https://github.com/tinylib/msgp/releases)

tianon:
Killer! :heart: :heart:

tianon:
FYI this fixed `armel` builds too: https://buildd.debian.org/status/fetch.php?pkg=golang-github-tinylib-msgp&arch=armhf&ver=1.0~alpha-2&stamp=1445642380 :metal:

tianon:
(available via https://packages.debian.org/sid/msgp and a mirror near you "soon" :+1:)

Issue #134, opened 2015-10-08
question: on uint vs. int decoding.


client9:
I started getting some of these errors on decode:

```
attempted to decode type "uint" with method for "int"
```

The culprit was another msgpack implementation

```
func (e *msgpackEncDriver) EncodeInt(i int64) {
    if i >= 0 {
        e.EncodeUint(uint64(i))
```

I found a few other libraries doing this trick.  Any suggestions here on how i can morph your library to read them?

update:

I'll spare you reading lua code, but  
https://github.com/fperrad/lua-MessagePack/

also defaults to unsigned.

```
set_integer'unsigned'
```

update, and one version of python, basically the pattern is "if non-negative, write as unsigned, if negative write as signed"

```

# You may notice struct.pack("B", obj) instead of the simpler chr(obj) in the
# code below. This is to allow for seamless Python 2 and 3 compatibility, as
# chr(obj) has a str return type instead of bytes in Python 3, and
# struct.pack(...) has the right return type in both versions.

def _pack_integer(obj, fp):
    if obj < 0:
        if obj >= -32:
            fp.write(struct.pack("b", obj))
        elif obj >= -2**(8-1):
            fp.write(b"\xd0" + struct.pack("b", obj))
        elif obj >= -2**(16-1):
            fp.write(b"\xd1" + struct.pack(">h", obj))
        elif obj >= -2**(32-1):
            fp.write(b"\xd2" + struct.pack(">i", obj))
        elif obj >= -2**(64-1):
            fp.write(b"\xd3" + struct.pack(">q", obj))
        else:
            raise UnsupportedTypeException("huge signed int")
    else:
        if obj <= 127:
            fp.write(struct.pack("B", obj))
        elif obj <= 2**8-1:
            fp.write(b"\xcc" + struct.pack("B", obj))
        elif obj <= 2**16-1:
            fp.write(b"\xcd" + struct.pack(">H", obj))
        elif obj <= 2**32-1:
            fp.write(b"\xce" + struct.pack(">I", obj))
        elif obj <= 2**64-1:
            fp.write(b"\xcf" + struct.pack(">Q", obj))
        else:
            raise UnsupportedTypeException("huge unsigned int")
```

client9:
Yarg, I forgot to say "thank you" in the first place for this project.  I love it, and thats why Im trying to figure out how to morph it to work with legacy applications.

[...]

PR #137, opened 2015-11-15
Сompatibility with msgppack-php

darvik80:
1. Changes in read*.go compatibility with https://github.com/msgpack/msgpack-php
2. Changes in gen/elem.go && gen/spec.go fix it:

```
//go:generate msgp
package main

type SomeStruct map[string]string

type SomeMegaStruct struct {
    Struct1 *SomeStruct `msg:"struct-1"`
    Struct2 *SomeStruct `msg:"struct-2"`
}
```

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/137)

### DIFF
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -317,12 +317,12 @@ func (s *Ptr) SetVarname(a string) {
 		if x.Value == IDENT {
 			x.SetVarname(a)
 		} else {
-			x.SetVarname("*" + a)
+			x.SetVarname("(*" + a+")")
 		}
 		return
 
 	default:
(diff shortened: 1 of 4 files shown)


philhofer:
The best way to read numbers that come from languages that do not draw distinctions between the signed and unsigned integers (or floating-point numbers, for that matter) is to use `msgp.Number`, which will happily decode signed and unsigned integers and single- and double-precision floats.

client9:
see also #134

shabbyrobe:
I think this is covered by #215 too?

PR #143, opened 2015-12-25
coerce numbers when appropriate

philhofer:
Addresses some comments in #134 
@client9

[Review on Reviewable](https://reviewable.io/reviews/tinylib/msgp/143)

### DIFF
--- a/msgp/number.go
+++ b/msgp/number.go
@@ -91,6 +91,70 @@ func (n *Number) Float() (float64, bool) {
 	}
 }
 
+// CoerceInt attempts to coerce the value of
+// the number into a signed integer, and returns
+// whether or not it was successful. "Success"
+// implies that no precision in the value of
+// the number was lost, which means that the
+// type of the number was an integer or 
+// a floating point 0.
+func (n *Number) CoerceInt() (int64, bool) {
+	switch n.typ {
+	case InvalidType, IntType:
+		return int64(n.bits), true
+	case UintType:
+		return int64(n.bits), n.bits <= math.MaxInt64
+	case Float32Type:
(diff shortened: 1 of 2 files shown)


klauspost:
@philhofer I got a bit nerd-sniped by this.

It seems like we can also return float values that map *exactly* to an integer - meaning a value that is exactly the integer without any rounding.

```go

// isExactInt will return true if the number represents an integer value.
// NaN, Inf returns false.
func (n *Number) isExactInt() bool {
	var eBits int // Exponent bits
	var mBits int // Mantissa bits

	switch n.typ {
	case InvalidType, IntType, UintType:
		return true
	case Float32Type:
		eBits = 8
		mBits = 23
	case Float64Type:
		eBits = 11
		mBits = 52
	default:
		return false
	}
	// Calculate float parts
	exp := int(n.bits>>mBits) & ((1 << eBits) - 1)
	mant := n.bits & ((1 << mBits) - 1)
	if exp == 0 && mant == 0 {
		// Handle zero value.
		return true
	}

	exp -= (1 << (eBits - 1)) - 1
	if exp < 0 || exp == 1<<(eBits-1) {
		// Negative exponent is never integer (except zero handled above)
		// Handles NaN (exp all 1s)
		return false
	}

	if exp >= mBits {
		// If we have more exponent than mantissa bits it is always an integer.
		return true
	}
	// Check if all bits below the exponent are zero.
	return bits.TrailingZeros64(mant) >= mBits-exp
}
```

It doesn't check if the value will under/overflow a given integer size, though.

klauspost:
Added in #404 with float coercion.

PR #148, opened 2016-03-27
support for aliased imports. Fixes #146.

### LINKED ISSUE #146: handling elements from another library - auto import copying
A fragment from replay.go, where I want to generate the msgp methods for the struct ReplayRequest:

```
import (
    tf "github.com/glycerine/tmframe"
)

//go:generate msgp                                                                                      

type ReplayRequest struct {
    Begin  *tf.Date
    Endx   *tf.Date
    Topics []string
}
```
where tf.Date is defined in a separate file:
https://github.com/glycerine/tmframe/blob/master/date.go#L13

and tf.Date already has the msgp method generation done for it.

Currently I have to hand edit the generated replay_gen.go to import tmframe.

Those hand edits will of course be lost when the file is re-generated.

It would be very helpful to have a way to tell msgp to (manually if not automatically) to import the packages it needs in turn in order to resolve the tf in tf.Date.

Or perhaps there is another way?

glycerine:
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/148)

### DIFF
--- a/parse/getast.go
+++ b/parse/getast.go
@@ -21,6 +21,7 @@ type FileSet struct {
 	Specs      map[string]ast.Expr // type specs in file
 	Identities map[string]gen.Elem // processed from specs
 	Directives []string            // raw preprocessor directives
+	Imports    []*ast.ImportSpec   // imports
 }
 
 // File parses a file at the relative path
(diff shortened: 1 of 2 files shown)


philhofer:
Nice. Thanks.

PR #153, opened 2016-04-15
implements cast methods for (u)int64 and (u)int32

dwlnetnl:
See discussion #134.

---

This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/153)

### DIFF
--- a/msgp/number.go
+++ b/msgp/number.go
@@ -91,6 +91,58 @@ func (n *Number) Float() (float64, bool) {
 	}
 }
 
+// CastInt64 returns the number as an int64 and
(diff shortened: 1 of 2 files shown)


philhofer:
You've added some additional same-type casts here, but I think #134 was more about cross-type casts that succeed because the 'value' of the number is the same (e.g., the bits `0x0000000000000000` and `0x8000000000000000` are both the value 0 under IEEE-754, so casting those to any integral type would be legal. Similarly, integral numbers that fit in a floating-point mantissa could be legally casted to floats as well.)

PR #156, opened 2016-05-11
[WIP] Avoid pointers receivers

### LINKED ISSUE #155: Use non-pointer receiver for Marshal and Size
I see from [this part of the code](https://github.com/tinylib/msgp/blob/master/gen/spec.go#L223) that the generator prefer pointer receivers for struct with > 3 fields and arrays.

This seems unnecessary (this methods do not modify the data they operate on, so they do not require a pointer) and is inconvenient (I'd like to be able to Marshal my structs without referencing them, which possibly increases the generated garbage).

Folks at Easyjson accepted [my PR](https://github.com/mailru/easyjson/pull/15) after checking benchmarks. Would you accept something similar for the msgp generator?.

Or is there some benchmark showing that the use of a pointer receiver actually improves performances?

hectorj:
Closes #155

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/156)

### DIFF
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -222,29 +222,7 @@ func next(t traversal, e Elem) {
 
 // possibly-immutable method receiver
 func imutMethodReceiver(p Elem) string {
-	switch e := p.(type) {
-	case *Struct:
-		// TODO(HACK): actually do real math here.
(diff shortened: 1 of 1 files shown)


hectorj:
Closing for now.

PR #157, opened 2016-05-23
gen: use msgp_ prefix on random variable names

alexandercampbell:
Fix a case I was seeing where `var` was generated as a "unique" variable name. Of course, the compiler complained about `var var`.

The issue wasn't easily reproduceable because it depends on `rand.Intn`, but fortunately, msgp doesn't seed the random generator and I was able to reproduce the issue enough times to find a fix.

@philhofer

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/157)


philhofer:
Thanks for the fix.

I think it might be a little friendlier to use `z{name}` instead of `msgp_{name}`, as it won't cause `golint` (and friends) to complain about underscores in identifiers.

alexandercampbell:
@philhofer good call. Addressed.

Issue #159, opened 2016-07-13
Ability to name struct fields as, e.g. uint16


gobwas:
Hello! 
Thank you for such a super fast library 🚀 

Let say we have a struct:
```go
type Pokemon struct {
    Name string
}
```
Generated code for which becomes:
```go
// somewhere in (*Pokemon) {Unmarshal,Decode}Msg
switch msgp.UnsafeString(field) {
		case "Name":
                // ...
```

Could it become somehow to something like this?
```go
type Pokemon struct {
    Name string `msg:"uint16,0x01"`
}
//...
// somewhere in (*Pokemon) {Unmarshal,Decode}Msg
switch binary.Uint16(field) {
		case 0x01:
                // ...

```

PR #160, opened 2016-07-15
Struct fields numeric labels

gobwas:
Hello! 

This pull request is about [this issue](https://github.com/tinylib/msgp/issues/159).

I think it could be useful. Some integration protocols use scheme, when map keys are numerical (`int` or `uint`), for example, to reduce network usage and other performance needs.

Regards,
Sergey.

### DIFF
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -315,7 +315,7 @@ func groupFieldsByType(fields []StructField) map[msgp.Type][]StructField {
 		switch f.FieldTag.(type) {
 		case int, int8, int16, int32, int64:
 			t = msgp.IntType
-		case uint, uint8, byte, uint16, uint32, uint64:
+		case uint, uint8, uint16, uint32, uint64:
 			t = msgp.UintType
 		case string:
 			t = msgp.StrType
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -281,6 +281,7 @@ func genStructFieldsParser(t traversalAssigner, p printer, fields []StructField)
 	hasUint := len(groups[msgp.UintType]) > 0
 	hasInt := len(groups[msgp.IntType]) > 0
 	hasStr := len(groups[msgp.StrType]) > 0
+	singleType := len(groups) == 1
 
 	if hasStr {
 		t.declareOnce(p, fieldBytes, "[]byte")
@@ -304,7 +305,7 @@ func genStructFieldsParser(t traversalAssigner, p printer, fields []StructField)
 	if hasInt {
 		t.declareOnce(p, fieldInt, "int64")
 	}
-	if hasInt || hasUint {
+	if !singleType || hasUint {
 		t.declareOnce(p, typ, "msgp.Type")
 	}
 
@@ -312,11 +313,14 @@ func genStructFieldsParser(t traversalAssigner, p printer, fields []StructField)
 	p.declare(sz, u32)
 	t.assignAndCheck(sz, mapHeader)
 	p.printf("\nfor %s > 0 {\n%s--", sz, sz)
-
-	if hasStr && !hasUint && !hasInt { // there are no numerically labeled fields, so parse every field label as string
+	switch {
+	case singleType && hasStr:
 		t.assignAndCheck(fieldBytes, mapKey)
 		switchFieldKeysStr(t, p, fields, fieldBytes)
-	} else {
+	case singleType && hasInt:
+		t.assignAndCheck(fieldInt, "Int64")
+		switchFieldKeys(t, p, fields, fieldInt)
+	default:
 		// switch on inferred type of next field
 		t.nextTypeAndCheck(typ)
 		p.printf("\nswitch %s {", typ)
@@ -340,7 +344,6 @@ func genStructFieldsParser(t traversalAssigner, p printer, fields []StructField)
 		t.skipAndCheck()
 		p.closeblock() // close switch
 	}
-
 	p.closeblock() // close loop
 }
 
@@ -372,10 +375,8 @@ func switchFieldKeysStr(t traversalAssigner, p printer, fields []StructField, la
 			return
 		}
 	}
-
 	p.print("\ndefault:")
 	t.skipAndCheck()
-
 	p.closeblock() // close switch
 }
 
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -389,10 +389,8 @@ func switchFieldKeys(t traversalAssigner, p printer, fields []StructField, label
 			return
 		}
 	}
-
 	p.print("\ndefault:")
 	t.skipAndCheck()
-
 	p.closeblock() // close switch
 }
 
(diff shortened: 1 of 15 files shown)


gobwas:
Also, I have a question about decision according to [this](https://github.com/tinylib/msgp/issues/134) thread. If it is necessary to declare label types also as `int32`, `int64`, `uint8` etc., I could implement it.

philhofer:
Hey, thanks for the PR.

This is going to take me a day or two to review, since it's a substantial change.

Just for my own edification, could you point me to a place where this would be used 'in the wild?'

philhofer:
Also, can I suggest the following for syntax:
```go
type Foo struct {
    Bar string `msg:uint8(0x03)"
}
```
We should be able to disambiguate string field names from integer field names simply by the presence of quotes.

gen/spec.go
```diff
@@ -199,6 +224,176 @@ type traversal interface {
...
+		var t msgp.Type
+		switch f.FieldTag.(type) {
+		case int, int8, int16, int32, int64:
+			t = msgp.IntType
+		case uint, uint8, uint16, uint32, uint64:
+			t = msgp.UintType
+		case string:
+			t = msgp.StrType
+		default:
+			panic(fmt.Errorf("field %q has unknown tag type %T", f.FieldName, f.FieldTag))
```
philhofer: Don't panic here; print an error and exit. Same goes for field label parsing.

gobwas:
Hey @philhofer,

Yeah, I agree with syntax you suggested. 

But my question above is still actual – should we take care on strong typed labels, like `uint8`? 
For example, we could use `uint` by default, and then add ability to cast not only to signed int, but to any other type of int? How do you think?

About usage 'in the wild' I do not clearly understand your question. If this is about concrete use case – I need this to implement some binary protocol to integrate with. It is very much like, for example, [this one](https://tarantool.org/doc/dev_guide/box-protocol.html).

philhofer:
> I need this to implement some binary protocol to integrate with. It is very much like, for example, this one.

Great; I was just curious about who was serializing things this way already.

> But my question above is still actual – should we take care on strong typed labels, like uint8?
For example, we could use uint by default, and then add ability to cast not only to signed int, but to any other type of int? How do you think?

Practically speaking, I'd imagine most things will end up being encoded as a 127-bit fixint anyway (and most everything else in one of the fixed-width unsigned types), so I don't think this is necessary, but I'm happy to be convinced otherwise.

gobwas:
Hello @philhofer!

Just returned to complete this feature. And please look what I have found about the use of field tag values without quotes: [play.golang.org](https://play.golang.org/p/MSsyJGYA8j). That is, it is not possible to distinguish numeric label vs. string simply by presence of quoutes. 
So I could suggest to use something like this (I decided that `(uint)0x01` syntax is bad because of super low, but probability of shadowing user key):

```go
type Foo struct {
    Bar string `msg:"0x03,uint"`
}
```

gen/spec.go
gobwas: There are already few cases when generator panics. For example, [here](https://github.com/tinylib/msgp/blob/master/gen/spec.go#L219). So is it really necessary to print error and do some of `os.Exit(1)`?

gen/spec.go
```diff
@@ -199,6 +224,181 @@ type traversal interface {
...
+		var t msgp.Type
+		switch f.FieldTag.(type) {
+		case int, int8, int16, int32, int64:
+			t = msgp.IntType
+		case uint, uint8, uint16, uint32, uint64:
+			t = msgp.UintType
+		case string:
+			t = msgp.StrType
+		default:
+			panic(fmt.Sprintf(
```
gobwas: Will duplicate my previous comment to not loose it:
> There are already few cases when generator panics. For example, here. So is it really necessary to print error and do some of os.Exit(1)? 

If it is, I could move log logic from `parse` package to separate package, e.g. `internal/log`, and reuse functions like `warnf`,`warnln` etc., and also refactor `fatalf` to print log message and do `os.Exit(1)`.

gobwas:
@philhofer ping 👀

parse/getast.go
```diff
@@ -339,14 +340,34 @@ func (fs *FileSet) getField(f *ast.Field) []gen.StructField {
...
+		}
+
+		if len(tags) > 1 {
+			last := len(tags) - 1
+			extension = tags[last] == "extension"
+
+			var err error
+			switch tags[1] {
+			case "uint":
+				str, base := numeric(tags[0])
```
philhofer: `strconv.ParseInt(tags[0], 0, 64)` will infer hex or octal based on the prefix; no need to do that yourself.

msgp/read.go
```diff
@@ -376,6 +376,7 @@ func (m *Reader) ReadMapKeyPtr() ([]byte, error) {
 			return nil, err
 		}
 		read = int(big.Uint32(p[1:]))
+
```
philhofer: dead extra line here

gen/spec.go
philhofer: That sounds great.

gen/spec.go
```diff
@@ -199,6 +224,181 @@ type traversal interface {
...
+		)
+		t.Fuse(data)
+
+		next(t, f.FieldElem)
+	}
+}
+
+func genStructFieldsParser(t traversalAssigner, p printer, fields []StructField) {
+	const (
+		fieldBytes = "field_b"
```
philhofer: Using `snake_case` will trip up tools like `golint`. It would be friendlier to use something more idiomatic.

gobwas:
@philhofer thanks! Applied your comments, and moved logging logic to the `internal/log` package. I did not changed already written `panic`'s to `log.Fatal`.

gobwas:
Hello @philhofer! Is it able to merge this? ) It becomes little bit uncomfortable to live with fork 😞

gobwas:
Ping @philhofer =)

gobwas:
I have also refactored code generation for maps – now you could marshal/unmarshal not only `map[string]T`.

gobwas:
@philhofer could you stop ignoring this? =)

pwaller:
FYI @gobwas, the tests failed (I'm not associated with this project, but just thought you might like to know):

```
--- FAIL: TestReadIntf (0.00s)
	read_test.go:63: map[thing-1:thing-1-value thing-2:800 thing-3:[115 111 109 101 32 105 110 110 101 114 32 98 121 116 101 115 46 46 46] thing-4:false] in; map[thing-4:false thing-1:thing-1-value thing-2:800 thing-3:[115 111 109 101 32 105 110 110 101 114 32 98 121 116 101 115 46 46 46]] out
FAIL
FAIL	github.com/tinylib/msgp/msgp	0.147s
```

gobwas:
@pwaller fixed.

gobwas:
@philhofer may be today? :rofl:

PotterDai:
@philhofer  Wondering if this PR will still be merged?

PR #161, opened 2016-07-15
Add comment to generated Msgsize function

pwaller:
This avoids the error message raised by golint:

> .Msgsize should have comment or be unexported

https://github.com/golang/lint/blob/c7bacac2b21ca01afa1dee0acf64df3ce047c28f/lint.go#L777

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/161)

### DIFF
--- a/gen/size.go
+++ b/gen/size.go
@@ -79,6 +79,8 @@ func (s *sizeGen) Execute(p Elem) error {
 		return nil
 	}
 
+	s.p.comment("Msgsize returns the number of bytes occupied by the serialized message")
+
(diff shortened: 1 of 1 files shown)


gen/size.go
```diff
@@ -79,6 +79,8 @@ func (s *sizeGen) Execute(p Elem) error {
 		return nil
 	}
 
+	s.p.comment("Msgsize returns the number of bytes occupied by the serialized message")
```
philhofer: Would you mind re-wording this slightly to indicate that `Msgsize()` is an upper-bound estimate, and not the exact number of bytes?

philhofer:
Thanks for this! Just one minor in-line comment.

pwaller:
Done.

PR #162, opened 2016-08-01
Make pointer-to-slice decoding work

pwaller:
Previously, the generated code was invalid because of the
operator precedence of * being lower than [.

The fix wraps the varname in parenthesis to avoid the problem.

This fixes the generated code for the following:

```go
package x

//go:generate msgp

type X struct {
	A int
}

type Y struct {
	Xs *[]X
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/162)

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -51,6 +51,7 @@ type TestType struct {
 	Num      msgp.Number `msg:"num"`
 	Slice1   []string
 	Slice2   []string
+	SlicePtr *[]string
 }
 
 //msgp:tuple Object
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -278,7 +278,12 @@ type Slice struct {
 func (s *Slice) SetVarname(a string) {
 	s.common.SetVarname(a)
 	s.Index = randIdent()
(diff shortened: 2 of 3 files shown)


pwaller:
For reference, here are the errors produced without this fix applied:

```
# github.com/tinylib/msgp/borked
./x_gen.go:138: cannot slice z.Xs (type *[]X)
./x_gen.go:156: invalid operation: z.Xs[zbai] (type *[]X does not support indexing)
./x_gen.go:204: invalid operation: z.Xs[zbai] (type *[]X does not support indexing)
./x_gen.go:227: invalid operation: z.Xs[zbai] (type *[]X does not support indexing)
./x_gen.go:266: cannot slice z.Xs (type *[]X)
./x_gen.go:284: invalid operation: z.Xs[zbai] (type *[]X does not support indexing)
```

philhofer:
Thanks. Mind adding a test case? (Adding a field to one of the structs in `_generated/def.go` is the easiest way to go about that.)

pwaller:
59b6b74 exercises the compile error without the fix applied. Is that enough, if not can you suggest what else?

philhofer:
Looks good to me!

PR #172, opened 2017-01-01
switch root element when inlining

### LINKED ISSUE #171: small recursive type leads to infinite loop
I found this when testing ZebraPack, it appears to also be present in the upstream msgp:  the small (low complexity) recursive struct Tree below causes an infinite loop during codegen. There is a check below in parse/inline.go (https://github.com/tinylib/msgp/blob/master/parse/inline.go#L118), but the types look different when printed so I don't think the check works anymore.

~~~
package main

//go:generate msgp

type S2 struct {
	A struct{}
	B string
	R map[string]uint8
	P uint16
	Q uint32
	T int64

	Arr [6]float64

	MyTree *Tree
}

type Tree struct {
	Chld []Tree
	Str  string
}

func main() {}
~~~

the fix (at least to me; please advise--it works to fix this example but will it miss other cases?) seemed something like this in parse/inline.go:

~~~
func (f *FileSet) nextInline(ref *gen.Elem, root string) {
    switch el := (*ref).(type) {
    case *gen.BaseElem:
        // ensure that we're not inlining                                                        
        // a type into itself                                                                    
        typ := el.TypeName()
        if el.Value == gen.IDENT && typ != root {
            if node, ok := f.Identities[typ]; ok && node.Complexity() < maxComplex {
+                // detect infinite loop:                                                         
+               if node.TypeName() == typ {
+                  infof("detected recursive type, not inlining %s\n", typ)
+                 return
+                }
~~~

philhofer:
In `nextInline`, inline on the copy of the replacement *after* the replacement has taken place, and use the replacement type name as the new root element.

@glycerine 

Fixes #171

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/172)

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -207,4 +207,5 @@ type Wrapper struct {
 type Tree struct {
 	Children []Tree
 	Element  int
+	Parent   *Wrapper
 }
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -199,3 +199,12 @@ type FileHandle struct {
 
 type CustomInt int
 type CustomBytes []byte
+
+type Wrapper struct {
+	Tree *Tree
+}
+
+type Tree struct {
+	Children []Tree
+	Element  int
+}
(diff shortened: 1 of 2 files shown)


glycerine:
Thanks @philhofer !

PR #181, opened 2017-06-27
no longer panics with the following file

karrick:
In `msgp/write.go`, the code that uses reflection to ensure the type of the map key is a string has a small bug.

Consider the following code sample, which ought to execute without panicking. After the fix, this code works as expected.

```Go
package main

//go:generate msgp

import (
	"bytes"
	"fmt"
	"os"

	"github.com/tinylib/msgp/msgp"
)

type Feedback map[string]interface{}

func main() {
	deepMap := make(map[string]map[string]string)
	feedback := Feedback{"general": deepMap}

	bb := new(bytes.Buffer)
	if err := msgp.Encode(bb, feedback); err != nil {
		fmt.Fprintf(os.Stderr, "%s\n", err)
		os.Exit(1)
	}
	fmt.Printf("%#v\n", bb.Bytes())
}

// 1. Type `go generate main.go`.
// 2. Type `go run main.go main_gen.go`.
```

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/181)

### DIFF
--- a/msgp/write.go
+++ b/msgp/write.go
@@ -689,7 +689,7 @@ func (mw *Writer) WriteIntf(v interface{}) error {
 }
 
 func (mw *Writer) writeMap(v reflect.Value) (err error) {
-	if v.Elem().Kind() != reflect.String {
+	if v.Type().Key().Kind() != reflect.String {
 		return errors.New("msgp: map keys must be strings")
 	}
 	ks := v.MapKeys()


philhofer:
Looks good. Thanks!

karrick:
I know this request seems trivial, but are you able to throw a release tag on master?

By the way, I just recently learned about tinylib/msgp and it's truly amazing! After spending a lot of time writing Avro libraries for Go, I truly think this is just one amazing library. More of our infrastructure will be moving to MessagePack because of this library. Thank you for all of your hard work!

philhofer:
Sure thing.

Issue #67, opened 2015-03-28
Let the shim functions return an error


zond:
At least the decode implementation usually has a few error cases, and it would be nice if those errors could be properly propagated.

philhofer:
Agreed

zond:
Do you want me to start hammering out a pull request, or will you do this? :)

philhofer:
I probably won't have time to look at this until later this week. Pull requests more than welcome.

zond:
I will try to take a look at it the next few days.

Any pointers on where to look?

philhofer:
[The rabbit hole starts here.](https://github.com/tinylib/msgp/blob/c8ed32c1281f411673da444ad77d3fe7fca408a2/parse/directives.go#L38)

philhofer:
It may make sense to implement this as a different directive. It may be tough to preserve backwards-compatibility otherwise.

zond:
I just ran out of time after fixing a bug in one of my projects, and then trying out your appengine build tag stuff. I will try to look at this during the weekend.

shabbyrobe:
@philhofer, would you be happy for me to pick this up? I really need this functionality too and I'm happy to do the work. Agree it should be a different directive.

philhofer:
Yeah, if you feel like taking a shot at this, go for it.

shabbyrobe:
OK I have a first pass of this, some feedback would be great! Ended up adding an optional extra argument to the directive instead of a new directive. The zero value defaults to the existing behaviour:

    //msgp:shim MyType as:string using:fromMyType/toMyType mode:convert

I ran into some sticky stuff with the size calculations. These shim functions could be doing all kinds of different stuff under the hood so it seems like it may be a performance tradeoff between running the shim function twice to get an accurate size or running it once and guessing the size, even if that guess will almost certainly be wrong in the case where the base type is a string. Either that or I'm missing something about how to do it more effectively... there's a good chance of that.

"Convert" might be a less than ideal name for it too as there is already a member on BaseElem called Convert, but I couldn't think of something better at the time.

shabbyrobe:
Reckon we can close this one off or does it still need a bit more work?

PR #182, opened 2017-07-17
First pass of #67 - Let the shim functions return an error

shabbyrobe:
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/182)

### DIFF
--- a/gen/decode.go
+++ b/gen/decode.go
@@ -143,6 +143,7 @@ func (d *decodeGen) gBase(b *BaseElem) {
 			d.p.printf("\n%s, err = dc.Read%s()", vname, bname)
 		}
 	}
+	d.p.print(errcheck)
 
 	// close block for 'tmp'
 	if b.Convert {
@@ -150,10 +151,9 @@ func (d *decodeGen) gBase(b *BaseElem) {
 			d.p.printf("\n%s = %s(%s)\n}", vname, b.FromBase(), tmp)
 		} else {
 			d.p.printf("\n%s, err = %s(%s)\n}", vname, b.FromBase(), tmp)
+			d.p.print(errcheck)
 		}
 	}
-
-	d.p.print(errcheck)
 }
 
 func (d *decodeGen) gMap(m *Map) {
--- a/gen/decode.go
+++ b/gen/decode.go
@@ -146,7 +146,11 @@ func (d *decodeGen) gBase(b *BaseElem) {
 
 	// close block for 'tmp'
 	if b.Convert {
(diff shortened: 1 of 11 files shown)


gen/decode.go
```diff
@@ -146,7 +146,11 @@ func (d *decodeGen) gBase(b *BaseElem) {
 
 	// close block for 'tmp'
 	if b.Convert {
-		d.p.printf("\n%s = %s(%s)\n}", vname, b.FromBase(), tmp)
+		if b.ShimMode == Cast {
+			d.p.printf("\n%s = %s(%s)\n}", vname, b.FromBase(), tmp)
+		} else {
+			d.p.printf("\n%s, err = %s(%s)\n}", vname, b.FromBase(), tmp)
```
philhofer: Here you're generating code that shadows `err` without checking it first.

gen/unmarshal.go
```diff
@@ -134,7 +134,12 @@ func (u *unmarshalGen) gBase(b *BaseElem) {
 	}
 	if b.Convert {
 		// close 'tmp' block
-		u.p.printf("\n%s = %s(%s)\n}", b.Varname(), b.FromBase(), refname)
+		if b.ShimMode == Cast {
+			u.p.printf("\n%s = %s(%s)\n", b.Varname(), b.FromBase(), refname)
+		} else {
+			u.p.printf("\n%s, err = %s(%s)\n", b.Varname(), b.FromBase(), refname)
```
philhofer: I think you've got the same `err` shadowing problem here.

philhofer:
Thanks for the PR.

I made a few inline comments, but overall this looks pretty good.

I wonder whether or not this should be a new directive, since `msgp:shim` is already pretty gross and verbose. Maybe `msgp:convert` or something like that.

shabbyrobe:
Thanks heaps for the comments, can't believe I missed that error thing 😳, should be fixed now.

I started out making it a separate directive but it shared so much of the same mechanics as `//msgp:shim`. I was also thinking that it might be possible to add support for #184 via the same "shim mode" mechanism (though I haven't really thought that through fully yet).

I'm not really fussed one way or the other. If you reckon it should be a separate directive, I'll dive back in and clean it up for you, no problem.

shabbyrobe:
Would you like me to update this to use a different directive?

philhofer:
Sorry for the slow review time. Looks Good. Thanks again for the PR.

shabbyrobe:
Awesome, thanks!

PR #188, opened 2017-08-10
Add support for rune

shabbyrobe:
I'm getting a `non-local identifier: rune` error when trying to generate a struct that has a `rune` as a field. This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/188)

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -1,9 +1,10 @@
 package _generated
 
 import (
-	"github.com/tinylib/msgp/msgp"
 	"os"
 	"time"
+
+	"github.com/tinylib/msgp/msgp"
 )
 
 //go:generate msgp -o generated.go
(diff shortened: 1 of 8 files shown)


philhofer:
The `ReadRune` situation is unfortunate. I wonder if we could work around it by having a built-in shim rule that knew `rune` was really just `int32`?

shabbyrobe:
I think the warning is just because of `bufio.Reader.ReadRune()`. The signature is `func (b *Reader) ReadRune() (r rune, size int, err error)`, which might not be the worst signature for it anyway... maybe that's one option?

Or I could take a crack at the built-in shim in a separate PR, maybe see which one looks like a better avenue from that angle?

philhofer:
If you're up to it, that would be great. (It's a protected function signature presumably because `ReadRune` should be an implementation of `io.RuneScanner`.)

 I already regret `(*Reader).ReadByte()`, but it's too late to fix that one.

shabbyrobe:
Closing in favour of #189

PR #189, opened 2017-08-16
Shim rune instead of WriteRune

shabbyrobe:
Here's v2 of the fix for handling runes. It works, it generates inline int32 casts just like it should, but the error still appears in the log:

    def.go: TestType: Rune: non-local identifier: rune
    def.go: TestType: RunePtr: non-local identifier: rune
    def.go: TestType: RunePtrPtr: non-local identifier: rune
    def.go: TestType: RuneSlice: non-local identifier: rune

I'll poke around now and see if I can silence the warning.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/189)

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -46,14 +46,19 @@ type TestType struct {
 		ValueA string `msg:"value_a"`
 		ValueB []byte `msg:"value_b"`
 	} `msg:"object"`
-	Child    *TestType   `msg:"child"`
-	Time     time.Time   `msg:"time"`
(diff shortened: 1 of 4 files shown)


shabbyrobe:
The same shimming could be used for time.Duration, which I always have to shim myself.

philhofer:
You may be able to silence the warning by always parsing `rune` as `int32`, since any `int32(rune(x))` type conversion is redundant. `time.Duration`, on the other hand, is a different type than `int64`.

PR #192, opened 2017-08-22
Fixes #191, panic in unsafe.UnsafeString()

### LINKED ISSUE #191: panic in unsafe.UnsafeString()
I came across a panic in `UnsafeString` yesterday while fuzzing another part of my app. I found a minimal reproducing test case:

```go
func TestUnsafe(t *testing.T) {
	var b = []byte{0xa0}
	o, _, _ := msgp.ReadStringZC(b)
	if msgp.UnsafeString(o) != "" {
		t.Fail()
	}
}
```

This panics because the length of the string is 0, so the use of `b[0]` in `unsafe.go` fails:

```
--- FAIL: TestUnsafe (0.00s)
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 5 [running]:
testing.tRunner.func1(0xc4200920f0)
	/usr/local/go/src/testing/testing.go:712 +0x2d2
panic(0x1137080, 0x1215f50)
	/usr/local/go/src/runtime/panic.go:491 +0x283
github.com/tinylib/msgp/msgp.UnsafeString(...)
	/Users/bl/code/go/src/github.com/tinylib/msgp/msgp/unsafe.go:27
```

The fix is trivial:

```go
func UnsafeString(b []byte) string {
	l := len(b)
	if l > 0 {
		return *(*string)(unsafe.Pointer(&reflect.StringHeader{Data: uintptr(unsafe.Pointer(&b[0])), Len: l}))
	} else {
		return ""
	}
}
```

It's highly unlikely this would ever be tripped by any generated code as it relies on a key of length 0 somehow making it past `ReadMapKeyPtr`, but it never hurts to remove this kind of potential from exported functions, you can never be sure everyone will heed the ALL CAPS EVIL WARNING!

shabbyrobe:
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/192)

### DIFF
--- a/msgp/unsafe.go
+++ b/msgp/unsafe.go
@@ -24,12 +24,8 @@ const (
 // THIS IS EVIL CODE.
 // YOU HAVE BEEN WARNED.
 func UnsafeString(b []byte) string {
-	l := len(b)
-	if l > 0 {
-		return *(*string)(unsafe.Pointer(&reflect.StringHeader{Data: uintptr(unsafe.Pointer(&b[0])), Len: l}))
-	} else {
-		return ""
-	}
+	sh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+	return *(*string)(unsafe.Pointer(&reflect.StringHeader{Data: sh.Data, Len: sh.Len}))
 }
 
 // UnsafeBytes returns the string as a byte slice
--- a/msgp/unsafe.go
+++ b/msgp/unsafe.go
@@ -24,7 +24,12 @@ const (
 // THIS IS EVIL CODE.
 // YOU HAVE BEEN WARNED.
 func UnsafeString(b []byte) string {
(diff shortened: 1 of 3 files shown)


msgp/unsafe.go
```diff
@@ -24,7 +24,12 @@ const (
 // THIS IS EVIL CODE.
 // YOU HAVE BEEN WARNED.
 func UnsafeString(b []byte) string {
-	return *(*string)(unsafe.Pointer(&reflect.StringHeader{Data: uintptr(unsafe.Pointer(&b[0])), Len: len(b)}))
+	l := len(b)
```
philhofer: You can avoid a branch by doing:

```go
sh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
return *(*string)(unsafe.Pointer(&reflect.StringHeader{Data: sh.Data, Len: sh.Len}))

```

shabbyrobe:
1.6ns per op vs 1.9, nice! Done.

PR #193, opened 2017-08-23
Fix unused variable in msgsize when using convert shimmode

shabbyrobe:
Found a bug in the implementation for #67 where an unused variable was causing the generated code to choke on compile. Added a few extra test cases.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/193)


philhofer:
LGTM! Thanks

Issue #184, opened 2017-07-17
Polymorphism support in generator


shabbyrobe:
I've been trying to find a neat way to deal with serialising a number of different types behind an interface. So far I've hit on two things that appear to work. One is to introduce an intermediate structure that has a 2-tuple of type/data, the other is to hack in a msgpack extension (which you explicitly recommend against in #35 - might be good to add that recommendation to the docs!).

In spite of the recommendation, the latter does indeed work, though not particularly well. It introduces a lot of intermediate copying of bytes and calculating `Len()` is difficult. I worked around that by marshalling in `Len()` and caching the result for the subsequent marshal, but you know... ew.

At the moment, I only have a few structures in my application that need this, but we are looking to serialise a heap more stuff so that's set to explode and a lot of the new stuff requires some sort of polymorhpism. I've been surviving by doing my own code generation of manually written msgp calls for the wrapper structs, but I'd really like a way to do this that doesn't require the intermediate struct because that will gunk up the application pretty badly.

I've been thinking something like another directive, like `//msgp:maptype ...`, which would  inline [type, data] as a 2-length array in place of the value itself. I'm imagining it could look something like this:

```go
//msgp:generate
//msgp:maptype Thing as:string using:thingToKind/kindToThing

func thingToKind(thing Thing) (kind string, err error) {
    switch thing.(type) {
    case ConcreteThing:
        kind = "concrete"
    default:
        err = fmt.Errorf("unknown thing %T", thing)
    }
    return
}

func kindToThing(kind string) (thing Thing, err error) {
    switch kind {
    case "concrete":
        thing = &ConcreteThing{}
    default:
        err = fmt.Errorf("unknown kind %s", kind)
    }
    return
}

type Thing interface {}
type ConcreteThing struct {
    Empty int
}
type Pants struct {
    Thing Thing
}
```

The mappers themselves are a bit annoying to write but it does seem to work well. I use ints for the kind but some may want to use strings.

```
// EncodeMsg implements msgp.Encodable
func (z *Msg) EncodeMsg(en *msgp.Writer) (err error) {
	var kind int
	var msg msgpsrv.Msg
	if z != nil {
		msg = z.Msg
		kind, err = mapper.Kind(msg)
		if err != nil {
			return
		}
	}

	// array header, size 2
	err = en.Append(0x92)
	if err != nil {
		return err
	}
	err = en.WriteInt(kind)
	if err != nil {
		return
	}
	if msg != nil {
[...]

Issue #35, opened 2014-11-29
Mechanism to generate Extension functions


dbudworth:
When one simply wants to create a set of custom types to be used as a field of type interface{}, you really just want to register a type number with a normal struct.

Currently, you must add 4 boiler plate functions to satisfy the Extension interface.

Is there a reason for making the extension interface not use the same function names as (Un)Marshaller and add the additional ExtensionType function?

Example code I have copy/pasted (and renamed struct) in my code:

```
func (i *Identity) ExtensionType() int8 {
    return IdentityT
}
func (i *Identity) Len() int {
    return i.Msgsize()
}
func (i *Identity) MarshalBinaryTo(b []byte) error {
    _, err := i.MarshalMsg(b)
    return err
}
func (i *Identity) UnmarshalBinary(b []byte) error {
    _, err := i.UnmarshalMsg(b)
    return err
}
```

PR #194, opened 2017-08-24
Implement polymorphism via msgp:intercept directive

shabbyrobe:
Here's an attempt at #184. I have had about 4 or 5 goes at this now, this is the only way I've found that went in even remotely cleanly. I tried doing it with extensions, then found the warning against that in #35. I tried extending ShimMode at least twice, but that was a total mess. This is the first time I've tried something that didn't feel like I was committing an act of vandalism!

Using this, we can delegate to a function to provide a stand-in that can decode, encode, marshal and unmarshal instead of calling it on the type itself. Interface types can be supported, as well as typed primitives.

There's no interface type for the mapper so we can retain type safety. You can just implement as much or as little as the code generator requires, based on what you ask it to generate. 

```go
//msgp:intercept MyIntf using:DefaultMyIntfMapper

type Thingy struct {
    Foo MyIntf
}

type MyIntf interface {}

func DefaultMyIntfMapper() myIntfMapper { return &myIntfMapper{} }

type myIntfMapper struct {}

func (p *myIntfMapper) DecodeMsg(dc *msgp.Reader) (t MyIntf, err error) {}
func (p *myIntfMapper) UnmarshalMsg(bts []byte) (t MyIntf, o []byte, err error) {}
func (p *myIntfMapper) EncodeMsg(t MyIntf, en *msgp.Writer) (err error) {}
func (p *myIntfMapper) MarshalMsg(t MyIntf, b []byte) (o []byte, err error) {}
func (p *myIntfMapper) Msgsize(t MyIntf) (s int) {}
```

The use is demonstrated in the `_generated/intercept_defs.go` and `_generated/intercept_test.go` files. I know I haven't got a complete set of tests here, but I need to think of more scenarios. I just thought I'd open this for discussion before I go test-case crazy just in case this approach is not one that would be acceptable.

I added a dummy element for interfaces that gets skipped by the printer but that was only to quiet the warning that gets raised: `intercept_defs.go: TestUsesIntfStructProvided: Foo: non-local identifier: TestIntfStructProvided`, if there's a better way to sort that out that involves fewer changes to the Elem hierarchy I'd be grateful for a tip, I didn't see one.

I know it looks like a big patch, but the overwhelming majority of it is tests!

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -194,7 +194,7 @@ type Custom struct {
 	Bts   CustomBytes          `msg:"bts"`
 	Mp    map[string]*Embedded `msg:"mp"`
 	Enums []MyEnum             `msg:"enums"` // test explicit enum shim
-	Some  FileHandle           `msg:file_handle`
+	Some  FileHandle           `msg:"file_handle"`
 }
 
 type Files []*os.File
--- a/_generated/intercept_defs.go
+++ b/_generated/intercept_defs.go
@@ -1,0 +1,257 @@
+package _generated
+
+import (
+	"fmt"
+
+	"github.com/tinylib/msgp/msgp"
+)
+
+//go:generate msgp
+
+//msgp:ignore testStructProvider
+type testStructProvider struct {
+	Events []string
+}
+
+var prv = &testStructProvider{}
+
+func resetStructProvider() {
+	prv = &testStructProvider{}
+}
+
+func TestStructProvider() *testStructProvider {
+	return prv
+}
+
+//msgp:intercept TestStructProvided using:TestStructProvider
+
+type TestStructProvided struct {
+	Foo string
+}
+
+type TestUsesStructProvided struct {
+	Foo *TestStructProvided
+}
+
+func (p *testStructProvider) DecodeMsg(dc *msgp.Reader) (t *TestStructProvided, err error) {
+	p.Events = append(p.Events, "decode")
+	t = new(TestStructProvided)
+	err = t.DecodeMsg(dc)
+	return
+}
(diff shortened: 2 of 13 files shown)


shabbyrobe:
Started dogfooding this in our project today, so far so good. Works seamlessly, supports all the interface types I have thrown at it, no panics. I've just realised JSON interop might require a bit more exploration though - that's not a feature we're using but it's a feature that msgp advertises so I will investigate the impact.

TCBeliever:
I was using msgp and be confused when using with interface.
This solutions looks like good. but it still in feature branch long time.
When does this feature merge to master?

shabbyrobe:
I've had this in production for close to a year now via a fork, it worked very well for our needs but it hasn't been vetted or reviewed by Phil. I'd be happy to help polish it up if there was something that needed to be done, but in the meantime if you want to make use of this you can make your own fork like I did and apply this PR to it.

shabbyrobe:
@philhofer been a long time since I followed up on this, my apologies. If the conflicts were resolved, do you see a pathway to merging this feature or should I close up my PR & fork?

klauspost:
Closing. Please reopen if you still think this is relevant

klauspost:
@shabbyrobe To be honest, I don't understand what it is doing. Isn't this just the [replace directive](https://github.com/tinylib/msgp/wiki/Preprocessor-Directives#replace)?

shabbyrobe:
No, this is more for serialising or deserialising an interface, which could be one of many concrete types. That looks like it lets you delegate to a different concrete struct.

PR #195, opened 2017-08-29
Fixed width array length constants other than uint32 can now compared to array headers

shabbyrobe:
I ran into a problem where using an integer width other than uint32 for array length constants was causing an error in the generator:

```go
const Bar int = 8
type Foo struct {
    Hello [Bar]string
}
```

This would yield errors like the following:

```
# github.com/tinylib/msgp/_generated
_generated/generated.go:1849:12: invalid operation: zigk != Bar (mismatched types uint32 and int)
_generated/generated.go:1850:33: cannot use Bar (type int) as type uint32 in field value
_generated/generated.go:1877:27: cannot use Bar (type int) as type uint32 in argument to en.WriteArrayHeader
_generated/generated.go:1896:28: cannot use Bar (type int) as type uint32 in argument to msgp.AppendArrayHeader
_generated/generated.go:1925:12: invalid operation: zuop != Bar (mismatched types uint32 and int)
_generated/generated.go:1926:33: cannot use Bar (type int) as type uint32 in field value
```

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/195)

### DIFF
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -3,7 +3,6 @@ package gen
 import (
 	"fmt"
 	"math/rand"
-	"strconv"
 	"strings"
 )
 
@@ -614,18 +613,6 @@ func writeStructFields(s []StructField, name string) {
 // can declare array lengths as any constant integer width, which breaks when
 // attempting a direct comparison to an array header's uint32.
 //
-// If we have a string at this point, it represents a constant which could have
-// any integer width so we must coerce it to uint32 to guarantee it is
-// comparable.
-//
 func coerceArraySize(asz string) string {
-	if _, err := strconv.ParseInt(asz, 10, 64); err == nil {
-		return asz
-	}
-	if strings.HasPrefix(asz, "0x") {
-		if _, err := strconv.ParseInt(asz[2:], 16, 64); err == nil {
-			return asz
-		}
-	}
 	return fmt.Sprintf("uint32(%s)", asz)
 }
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -3,6 +3,7 @@ package gen
 import (
 	"fmt"
 	"math/rand"
+	"strconv"
 	"strings"
 )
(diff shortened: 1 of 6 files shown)


gen/elem.go
```diff
@@ -605,3 +606,21 @@ func writeStructFields(s []StructField, name string) {
...
+// ArrayHeader implementation in this library using uint32. On the Go side, we
+// can declare array lengths as any constant integer width, which breaks when
+// attempting a direct comparison to an array header's uint32.
+//
+// If we have a string at this point, it represents a constant which could have
+// any integer width so we must coerce it to uint32 to guarantee it is
+// comparable.
+//
+func coerceArraySize(asz string) string {
+	if _, err := strconv.ParseInt(asz, 10, 64); err != nil {
```
philhofer: Typed constants can be base 16, too. Or base 8.

It is probably more robust to see if the constant is typed when parsing the AST.

shabbyrobe: Ah cool, will have another dip tomorrow, see how it looks. Update this PR or submit a new one?

shabbyrobe: If I can't find a neat way to do that during the AST parse, this could just check if the first character is a digit, or check if the token is a valid identifier and cover those cases too

shabbyrobe: But that is definitely a bit more hacky than using the AST, if possible

shabbyrobe:
I've fixed this solution to work with hex and octal as I need it in my own fork, I'll have a go at an AST based solution today or tomorrow.

philhofer:
Another alternative is just to always cast to `uint32`. That's probably the simplest solution.

shabbyrobe:
That'd be the simplest solution by far. I don't know why I was under the impression that casts had a runtime cost but I did a quick benchmark and it looks like they don't. I'll update now.

PR #204, opened 2017-09-25
use the appropriate base conversion in decode generation

### LINKED ISSUE #203: Several Generation Issues
I'm having issues where the code being generated is rather error-filled. Missing elses and forced conversions when shimming are the primary issues.

As an example:

https://gist.github.com/ColtonProvias/c5a020bb8ba42942689d114488568354

Line 24 in sample_gen.go attempts to force a conversion that throws an error during compilation. And just 2 lines above an else is missing in the generated code.

philhofer:
Fixes #203.

### DIFF
--- a/gen/decode.go
+++ b/gen/decode.go
@@ -128,7 +128,7 @@ func (d *decodeGen) gBase(b *BaseElem) {
 	switch b.Value {
 	case Bytes:
 		if b.Convert {
-			d.p.printf("\n%s, err = dc.ReadBytes([]byte(%s))", tmp, vname)
+			d.p.printf("\n%s, err = dc.ReadBytes(%s)", tmp, tobaseConvert(b))
 		} else {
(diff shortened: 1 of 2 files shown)


gen/decode.go
```diff
@@ -128,7 +128,7 @@ func (d *decodeGen) gBase(b *BaseElem) {
 	switch b.Value {
 	case Bytes:
 		if b.Convert {
-			d.p.printf("\n%s, err = dc.ReadBytes([]byte(%s))", tmp, vname)
+			d.p.printf("\n%s, err = dc.ReadBytes(%s)", tmp, tobaseConvert(b))
```
ColtonProvias: For convert mode, the functions being called return two values `([]byte, error)`, so I think this will throw an error.

klauspost:
Obsolete

PR #205, opened 2017-10-03
Field context information in error messages

shabbyrobe:
We have a lot of trouble debugging our server when client applications fail to encode fields correctly. I have prepared a patch that adds information about the problematic field to the error output. Benchmarks don't seem to be adversely affected. It looks like a big patch but it's mostly tests and additions to the error types.

I think I have managed to do it without breaking backward compatibility with the error types. The only compromise I can see is that ErrShortBytes can not include context information without introducing such a break.

I haven't quite worked out how best to test writer failures yet during Marshal or Encode.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/205)

### DIFF
--- a/_generated/convert_test.go
+++ b/_generated/convert_test.go
@@ -12,8 +12,8 @@ func TestConvertFromEncodeError(t *testing.T) {
 	var buf bytes.Buffer
 	w := msgp.NewWriter(&buf)
 	err := e.EncodeMsg(w)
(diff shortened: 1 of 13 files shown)


philhofer:
I rebased this patch series on master and merged. Thanks!

shabbyrobe:
Awesome, thank you!

PR #206, opened 2017-10-10
WIP: Add omitempty annotation support

bozaro:
**This PR is incomplete becuase it don't contains new tests and most likely contains contentious issues. 
I will be glad to receive feedback to make this changes production-ready.**

Reason: add ability to skip empty fields on serialization (like default `json`/`xml` marshaling) for reduce serialized object size and deserialization time.

Changes summary:

 * add "omitempty" option specifies: the field should be omitted from the encoding if the field has an empty value;
 * add autogenerated method `ShouldOmitAsEmpty() bool` for structures;
 * add methods `msgp.ShouldOmitAsEmpty<base type>(v) bool` for primitive types;
 * add flag `-omitempty` for `msgp` generator (default: `true`).

Field omit rules (only for marked fields):

 * struct - if method `ShouldOmitAsEmpty()` returns `true` (generated method return `true` if all serializable fields can be ommited);
 * pointer - on `nil` value;
 * fixed size array - newer;
 * slice - on empty size;
 * map - on `nil` or empty.

Incompatible changes:

 * Generated `EncodeMsg`, `MarshalMsg` code try invoke `ShouldOmitAsEmpty` method for structures with custom marshaling. Can be disabled with `-omitempty=false` flag.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/206)

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -61,6 +61,34 @@ type TestType struct {
 	SlicePtr   *[]string
 }
 
+type TestOmit struct {
+	F   *float64          `msg:"float,omitempty"`
+	Els map[string]string `msg:"elements,omitempty"`
+	Obj struct {          // test anonymous struct
+		ValueA string `msg:"value_a,omitempty"`
+		ValueB []byte `msg:"value_b,omitempty"`
+	} `msg:"object,omitempty"`
+	Child      *TestOmit   `msg:"child,omitempty"`
+	Time       time.Time   `msg:"time,omitempty"`
+	Any        interface{} `msg:"any,omitempty"`
+	Appended   msgp.Raw    `msg:"appended,omitempty"`
+	Num        msgp.Number `msg:"num,omitempty"`
+	Byte       byte        `msg:"byte,omitempty"`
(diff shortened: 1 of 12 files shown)


erikdubbelboer:
This core runs into the same issue as the previous time someone made a pull request for this: [This code would make reusing structs impossible.](https://github.com/tinylib/msgp/pull/164#issuecomment-254035661)

bozaro:
I do not see this as a problem:

 * No one forces `omitempty` usage;
 * No one guaranteed the possibility of reusing structures.

For reusing structs:

 * You must not add `omitempty` option (this PR don't force omitempty);
 * You must ensure, that all objects contains all fields (in our case we sometime run two versions on same database).

pavel-odintsov:
Hello!

Can I ask about any updates regarding this task? That's quite important thing to have.

Thank you!

bradleypeabody:
I'm pretty sure this functionality has since been fully addressed with later changes.  `omitempty` support was merged a while ago, and `omitzero` provides the functionality of allowing structs or any custom type to omit itself.

klauspost:
Agree.

PR #209, opened 2017-12-07
Removed unnecessary "err" of naked return

dchenk:
The generated encode functions that have a naked return (simply "return") where the function returns an error type all have the "err" variable declared as a named return (as it is here in the func "appendraw"). But this is the only place where the "return" is followed by a " err"; this is unnecessary because the return value is named in the function signature. I've been manually removing the extra space-err in every .Append block because they make the code inconsistent and are very annoying.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/209)

### DIFF
--- a/gen/encode.go
+++ b/gen/encode.go
@@ -101,7 +101,7 @@ func (e *encodeGen) appendraw(bts []byte) {
 		}
 		e.p.printf("0x%x", b)
 	}
-	e.p.print(")\nif err != nil { return err }")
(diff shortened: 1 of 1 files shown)


shabbyrobe:
Are there any instances at all in the generated code where `err` could be shadowed?

dchenk:
No shadowing is possible because the `return` immediately follows and the check for `err != nil` after the call to `.Append`

PR #210, opened 2017-12-10
Various cleanups to generator for consistency

dchenk:
It is more idiomatic in Go and cleaner code to simply have one variable set (without the blank identifier) when ranging over a map if only the key of each element is desired.

According to the language specification (here: https://golang.org/ref/spec#RangeClause):

> As with an assignment, if present the operands on the left must be addressable or map index expressions; they denote the iteration variables. If the range expression is a channel, at most one iteration variable is permitted, otherwise there may be up to two. If the last iteration variable is the blank identifier, the range clause is equivalent to the same clause without that identifier.

(Also here I'm adding two items to .gitignore—I would like to contribute further to this project.)

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/210)

### DIFF
--- a/parse/getast.go
+++ b/parse/getast.go
@@ -93,14 +93,14 @@ func File(name string, unexported bool) (*FileSet, error) {
 // applyDirectives applies all of the directives that
 // are known to the parser. additional method-specific
 // directives remain in f.Directives
-func (f *FileSet) applyDirectives() {
-	newdirs := make([]string, 0, len(f.Directives))
-	for _, d := range f.Directives {
+func (fs *FileSet) applyDirectives() {
+	newdirs := make([]string, 0, len(fs.Directives))
+	for _, d := range fs.Directives {
 		chunks := strings.Split(d, " ")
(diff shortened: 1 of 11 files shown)


parse/getast.go
```diff
@@ -136,21 +136,21 @@ func (f *FileSet) applyDirectives() {
 // figuring out that D is just a uint64.
 type linkset map[string]*gen.BaseElem
 
-func (f *FileSet) resolve(ls linkset) {
+func (fs *FileSet) resolve(ls linkset) {
 	progress := true
 	for progress && len(ls) > 0 {
 		progress = false
 		for name, elem := range ls {
-			real, ok := f.Identities[elem.TypeName()]
```
dchenk: "real" is an identifier of a built-in function in Go.

philhofer:
Please don't continue to push to branches for which there are merged commits.

Also, please create one PR per patch, and solve one issue per patch, so that I can review each of them separately.

Thanks.

PR #211, opened 2017-12-11
Added two lines to gitignore

dchenk:
I'd like to contribute more, so please accept. :)

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/211)


shabbyrobe:
@dchenk - this looks like it still has the extra commits in it.

You can also add things to your own global gitignore without necessarily adding them to the project. Add this to your `~/.gitconfig`:

```
[core]
	excludesfile = ~/.gitignore
```

And then you can add anything you like to your own `~/.gitignore`, which will apply to all projects.

dchenk:
Hi @shabbyrobe 
I guess I can just use my own global gitignore. But I'm actually not sure why the other commits are appearing here (same is happening in my other pull requests); any ideas?

PR #212, opened 2017-12-11
Added type name regexp matching

dchenk:
You can now use regular expressions for type names in directives. Instead of the concrete type name, you can prefix "reg=" before the regular expression to match, or prefix "reg!=" before the regular expression that the type names should NOT match.

This comes in extremely useful when you want to ignore a set of types that begin a particular way.

For example, this set of directives:
```
//go:generate msgp
//msgp:encode ignore reg=Req
//msgp:unmarshal ignore reg!=^Contents
```
indicate that types that include "Req" in the name will not get an `EncodeMsg` method, and only types that begin with "Contents" will get an `UnmarshalMsg` method.

You can use plain concrete type names as before, even along with regexp:
`//msgp:encode ignore reg=Req AnotherType`

Using a flag like `-io` or `-marshal` may not be enough for you if, for example, in general you don't want methods of a particular kind but do want it for only certain types. I'm finding regexp extremely useful for my work.

The regular expression syntax is described here: https://godoc.org/regexp

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/212)

### DIFF
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -3,6 +3,7 @@ package gen
 import (
 	"fmt"
 	"io"
+	"regexp"
 )
 
 const (
@@ -136,13 +137,37 @@ type TransformPass func(Elem) Elem
 
 // IgnoreTypename is a pass that just ignores
 // types of a given name.
-func IgnoreTypename(name string) TransformPass {
+func IgnoreTypename(pattern string) TransformPass {
 	return func(e Elem) Elem {
-		if e.TypeName() == name {
+		// Check if the type name is a regexp.
+		if TypeNameMatches(pattern, e.TypeName()) {
 			return nil
 		}
 		return e
 	}
+}
+
+// TypeNameMatches compares the pattern to the typeName to see if the type
+// name satisfies the pattern. The pattern can be either simply the type
+// name itself or a regexp pattern. A regexp pattern is extracted by taking
+// everything that follows either "reg=" or "reg!=" in the pattern string.
+// Pattern "reg=expr" returns true if and only if typeName matches expr.
+// Pattern "reg!=expr" returns true if and only if typeName does NOT match expr.
+func TypeNameMatches(pattern, typeName string) bool {
+	if len(pattern) > 4 && pattern[:3] == "reg" {
+		if string(pattern[3]) == "!" {
+			if !regexp.MustCompile(pattern[5:]).MatchString(typeName) {
+				fmt.Printf("Matched negated regexp %q to type name %q\n", pattern[5:], typeName)
+				return true
+			}
+		} else if regexp.MustCompile(pattern[4:]).MatchString(typeName) {
+			fmt.Printf("Matched regexp %q to type name %q\n", pattern[4:], typeName)
+			return true
+		}
+		return false
+	}
+	// Concrete type names compare by simple equality:
+	return pattern == typeName
 }
 
 // ApplyDirective applies a directive to a named pass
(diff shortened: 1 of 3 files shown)


dchenk:
not sure why other commits appear here, but they're all consistent with master

dchenk:
Hint: When using `reg!=` (to ignore all types whose name does not satisfy the regular expression), if you want generated methods for more than one type but all the others ignored, typically what you'll want to do is have one `reg!=` expression where the types to _include_ are separated by the regexp `|` operator. So, for example, you can write:
`//msgp:decode ignore reg!=FirstType|^Something|Another`
That tells the preprocessor to ignore all types except for the ones that satisfy at least one those three nested regular expressions (separated by the OR operator).

If you have this: 
`//msgp:decode ignore reg!=FirstType reg=FirstTypeOther`
then `encode` methods will be generated only for (a) the types whose names _do_ match the expression `FirstType` and (b) the types whose names do _not_ match the expression `FirstTypeOther`.

philhofer:
The old commits are still in this branch because you haven't rebased this branch against the new master.

dchenk:
Ok I see, thanks. Is there anything I need to do for us to get this pull request merged?

philhofer:
Yes; please rebase this branch and squash the fixup commits.

Some high-level comments:

 - New features need new regression tests so that I can't accidentally break this functionality and have tests continue to pass. In terms of checking for the presence or absence of text in the output of the tools, check out how `_generated/issue98.sh` works.
 - I'm not sure I like shoehorning new functionality into existing directives. Adding new directives keeps us from accidentally breaking old ones. So, if we decide this functionality is necessary, I'd rather stick it in a directive called something like `//msgp:match ...`
 - What does this feature accomplish that can't be accomplished by sticking the right types into the right files? (The biggest/only reason we have `msgp:ignore` is that running the tool with `-unexported` grabs literally everything. We didn't have a way to ignore types when the only behavior was to generate code for exported types.) If we're going to merge new features, let's make sure we understand what problem we're solving, rather than going for the first thing we come up with. I almost never _remove_ features, so please understand my reticence in accepting (and maintaining, in perpetuity) new ones.
 - For the most part, [I like my identifiers named the way they are](http://www.literateprogramming.com/pikestyle.pdf). Try to keep diffs minimal. As a general rule of thumb (not just for this project), smaller diffs get merged faster.

dchenk:
Ok I can't figure out how to remove commit 35bef8c from the history of this branch. But it's already merged into master, so only the three files changed in commit 9909a36 are being changed in this PR.

dchenk:
@philhofer Thank you for your thorough explanation. I would like to offer my help maintaining this repo going forward. So please don't worry about the maintainability of new features—I can try to help as much as you'd like. I'm still trying to fully understand how all the parts are put together. I'm also noticing some infelicities in the way the program is split up into the `gen` and `parse` packages—they don't exactly split up the work consistently or logically (for example where type names are matched to the names listed in directives). I may soon propose a few re-organizations of code to help with maintainability.

About this particular feature, I wouldn't say it involves "shoehorning new functionality into existing directives" because it's only adding a new way to _select types_ by name for a directive. I'm finding regexp extremely helpful for myself, where I just need certain methods (either to encode, or decode, etc.) for most types. Again, everything about the code generator continues to work as before for the people who may not need regular expressions.

About adding tests, I'll do my best to figure out how testing is set up in this project (though I can't find the file at `_generated/issue98.sh` that you mentioned). Do you mind at least helping me get started writing some tests?

Thanks a lot!

PR #215, opened 2018-01-29
Fix int/uint interoperation to conform to de-facto expectations of other clients

### LINKED ISSUE #134: question: on uint vs. int decoding.

shabbyrobe:
Been running this in production for a while now. This fixes #134 to meet the behaviour of other clients, which encode non-negative integers as unsigned regardless of whether the originating type is signed.

It raises an error on underflow or overflow and is tested pretty thoroughly (though I'm always happy to add more).

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/215)


dchenk:
I like this, but I don't see why ReadUint64 needs to be extended to signed integers. It's unlikely that feature will ever come in useful to anyone.

shabbyrobe:
If it doesn't hurt, and it works, why not? The spec is totally unclear about this one way or the other. What's to stop some client from implementing it the other way, and who would we be to say they were wrong?

msgp/read_bytes_test.go
```diff
@@ -222,53 +225,211 @@ func BenchmarkReadBoolBytes(b *testing.B) {
...
+		return false
+	}
+
+	vs := []struct {
+		v        interface{}
+		rdBits   int
+		failBits int
+		errCheck func(err error, failBits int) bool
+	}{
+		{math.MaxInt64, i32, 32, overflowErr},
```
philhofer: these need explicit casts in order to make the 386 build pass

shabbyrobe:
Oops, yep, sorry about that. I've fixed it up and it's passing on my machine with `GOARCH=386`, hopefully Travis doesn't spit chips!

PR #216, opened 2018-02-12
add missing magic number when map is sized from mm16 -> mm32

rakitzis:
Found this while using resizeMap for my own purposes; it doesn't fix a bug in msgp as `delta` is only set to -1 in the code I see. However since `resizeMap` plainly tries to do the right thing when sizing upwards it seems worth supplying a fix.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/216)

### DIFF
--- a/msgp/edit.go
+++ b/msgp/edit.go
@@ -198,6 +198,7 @@ func resizeMap(raw []byte, delta int64) []byte {
 		if cap(raw)-len(raw) >= 2 {
 			raw = raw[0 : len(raw)+2]
 			copy(raw[5:], raw[3:])
+			raw[0] = mmap32
 			big.PutUint32(raw[1:], uint32(sz+delta))
 			return raw
 		}


rakitzis:
NB: would be awesome if `resizeMap` were exported.

philhofer:
Nice.

I think the API would have to be re-thought for this to be exported. As it stands right now, it's a bit of a foot-gun. I've generally tried to steer clear from APIs that let you accidentally produce objects that can't be parsed.

Maybe an API for interleaving two maps (assuming that's what you're doing)? It'd be easy enough to make that streaming as well.

rakitzis:
Yeah, I am using this to append one map to another; the last entry "wins" and I do not bother filtering duplicates. The API matters -- I also care to append key/value pairs (just a `[]string` of pairs) which may not necessarily be a msgp-serialized map. That is a relatively minor concern, but having the sharp tool exposed lets me do exactly what I want to do.

PR #218, opened 2018-02-14
Support GopherJS and other pure Go environments

damz:
Unsafe pointer arithmetic is not supported under GopherJS, and currently fails silently.

Add the `purego` tag, standardized as part of golang/go#23172, in addition to the `appengine` tag kept for backwards compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/218)

### DIFF
--- a/msgp/purego.go
+++ b/msgp/purego.go
@@ -1,4 +1,4 @@
-// +build appengine
+// +build purego appengine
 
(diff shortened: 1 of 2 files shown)


damz:
@philhofer Thanks :)

PR #236, opened 2018-09-05
Default timestamp timezone to UTC rather than Local

tkuhlman:
I found it unintuitive that a timestamp is returned in my local timezone when the original timezone is not known (being lost in the serialization), rather returning in UTC seems better. This also makes the output consistent no matter where the machine happen to think it is located.

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/236)

### DIFF
--- a/_generated/convert_test.go
+++ b/_generated/convert_test.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"testing"
 
-	"github.com/tinylib/msgp/msgp"
+	"github.com/GannettDigital/msgp/msgp"
 )
 
 func TestConvertFromEncodeError(t *testing.T) {
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -4,7 +4,7 @@ import (
 	"os"
 	"time"
 
-	"github.com/tinylib/msgp/msgp"
+	"github.com/GannettDigital/msgp/msgp"
 )
 
 //go:generate msgp -o generated.go
--- a/_generated/def_test.go
+++ b/_generated/def_test.go
@@ -5,7 +5,7 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/tinylib/msgp/msgp"
+	"github.com/GannettDigital/msgp/msgp"
 )
 
 func TestRuneEncodeDecode(t *testing.T) {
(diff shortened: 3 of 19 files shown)


philhofer:
The code's current behavior guarantees that serializing the result of `time.Now()` (and friends) and then de-serializing the result will give you a `time.Time` value that passes a `==` comparison test with the input. This change breaks that behavior (but doesn't break well-behaved code that uses `(time.Time).Equal()` like it should be).

It's perfectly reasonable to argue that no one should be comparing `time.Time` values with `==`,
but I need a really good reason to break the existing API, particularly behavior that is
explicitly documented.

tkuhlman:
The statement that the code's current behavior guarantees that serializing the result of time.Now() and de-serializing will be `==` comparable is only true assuming that the machine serializing and the machine de-serializing are in the same time zone. The problem I am encountering is that I run a decentralized system and different parts of the system are in different timezones and because of that different parts of the system produce inconsistent results, in other words across the system results are not `==` comparable. In addition since there is no way to know what the de-serialized time zone will be ahead of time I can't simply change the input timezone to match. Instead the alternative is to change the output timezone, rather than change it in many places throughout my code it makes sense to do it once on de-serialization hence the PR.

erikdubbelboer:
@tkuhlman I suggest doing everything in UTC and only add time zones once you need to display something to a user. If you have control over the servers have them all in UTC as well. The amount of issues you will avoid this way is huge!

tkuhlman:
@erikdubbelboer I agree it is all easier if all servers are in UTC and all time stamps remain in UTC with time zones only added when displayed. That is what I am trying to accomplish. The catch in my case is I don't necessarily always have control of the timezone on the machine when de-serialization happens which is why having the de-serialization set it to UTC rather than local timezone is helpful. The timezone at the consumer end can still be reset later in either case but the layer of the service I control has passed it off to other code by then.

udryan10:
Making sure your systems are running in UTC is certainly a good start, but I don't think thats a "fix" per-se. These are not always headless systems, often times they are developer machines running tests.

klauspost:
A (de)serializer should not change the timezone. While convenient in some cases it would be destructive in others.

I know your pain and have several ugly "set timezone to UTC in all fields of this struct" using reflection, but it is simply not the job of a serializer.

PR #244, opened 2019-01-02
Fix bin encoded map keys

### LINKED ISSUE #243: Fail to decode string encoded as bin type
**Problem** 
msgp fails to decode string which were encoded as 'bin' type. It fails with the error: *msgp: too few bytes left to read object*

**Steps to reproduce:**

The following program generates the said error.

Note: The byte buffer is a valid msgpack buffer generated using embedded python program. Notice the use of "use_bin_type=True". When false, this error does not occur

```go
package main

import (
	"fmt"
	"os"
)


var buf = []byte{130, 196, 1, 97, 1, 196, 1, 98, 2}
// The above byte buffer was generated using the following python program
/*
import msgpack

m = {
		'a': 1,
		'b': 2,
}

buf = msgpack.dumps(m, use_bin_type=True)
print "[]byte{%s}" % ",".join([str(ord(n)) for n in buf])
*/

//go:generate msgp

type M struct {
	A int `msgp:"a"`
	B int `msgp:"b"`
}

func main() {
	m := &M{}
	_, err := m.UnmarshalMsg(buf[:])
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
	fmt.Printf("%v\n", m)
}

```

erikdubbelboer:
When returning a TypeError the caller may try to continue reading a different type from the returned buffer. So make sure we actually return the buffer and not nil.

Fixes #243

---
This change is [Reviewable](https://reviewable.io/reviews/tinylib/msgp/244)

### DIFF
--- a/msgp/read_bytes.go
+++ b/msgp/read_bytes.go
@@ -314,8 +314,7 @@ func ReadFloat32Bytes(b []byte) (f float32, o []byte, err error) {
 	}
 
 	if b[0] != mfloat32 {
-		err = TypeError{Method: Float32Type, Encoded: getType(b[0])}
-		return
+		return 0, b, TypeError{Method: Float32Type, Encoded: getType(b[0])}
 	}
 
 	f = math.Float32frombits(getMuint32(b))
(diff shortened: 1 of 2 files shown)


philhofer:
Thanks for the PR.

I agree that this fixes the issue, but I'd rather that the API to `ReadStringZC` remain the same.

The real bug is that `ReadMapKeyZC` shadows the local variable `b` on the first line of the function, so it erroneously tries to re-parse the data from a buffer that is always empty. That variable should be renamed.

erikdubbelboer:
Close in favor of https://github.com/tinylib/msgp/pull/245

PR #245, opened 2019-01-03
msgp/msgp: fix ReadMapKeyZC

### LINKED ISSUE #243: Fail to decode string encoded as bin type

philhofer:
Due to some unintentional variable shadowing, ReadMapKeyZC
never actually accepted 'bin'-typed map keys. Fix the shadowing
and add test coverage.

Fixes #243 

cc @erikdubbelboer @sudeep9

### DIFF
--- a/msgp/read_bytes.go
+++ b/msgp/read_bytes.go
@@ -201,14 +201,14 @@ func ReadMapHeaderBytes(b []byte) (sz uint32, o []byte, err error) {
 // - ErrShortBytes (too few bytes)
 // - TypeError{} (not a str or bin)
 func ReadMapKeyZC(b []byte) ([]byte, []byte, error) {
-	o, b, err := ReadStringZC(b)
+	o, x, err := ReadStringZC(b)
 	if err != nil {
 		if tperr, ok := err.(TypeError); ok && tperr.Encoded == BinType {
 			return ReadBytesZC(b)
 		}
 		return nil, b, err
 	}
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Sorry I haven't had time to update my pull request yet. This looks good 👍

sudeep9:
Looks ok to me 👍

PR #249, opened 2019-04-30
issue#163 Add getFieldsFromEmbeddedStruct function.

racketprogram:
```go
type Common struct {
	RequestID msgp.Number `msg:"i"`
}

type GetUserRequest struct {
	Common `msg:",flatten"`
	UserID uint32 `msg:"a"`
}
```


philhofer:
Thanks for the PR.;

Please add a test for this functionality in `_generated` so that I (or someone else) does not break it further down the road.

parse/getast.go
```diff
@@ -345,6 +346,9 @@ func (fs *FileSet) getField(f *ast.Field) []gen.StructField {
 		if len(tags) == 2 && tags[1] == "extension" {
```
philhofer: It might make sense to refactor this to
```go
if len(tags) >= 2 {
    switch tags[1] {
    case "extension": ...
    case "flatten": ...
    }
}
```

parse/getast.go
```diff
@@ -402,6 +410,22 @@ func (fs *FileSet) getField(f *ast.Field) []gen.StructField {
 	return sf
 }
 
+func (fs *FileSet) getFieldsFromEmbeddedStruct(f ast.Expr) []gen.StructField {
+	switch f := f.(type) {
+	case *ast.Ident:
+		s := fs.Specs[f.Name]
+		switch s := s.(type) {
```
philhofer: We should be able to handle bare identifiers here as structs.

parse/getast.go
racketprogram: done

racketprogram:
> Thanks for the PR.;
> 
> Please add a test for this functionality in `_generated` so that I (or someone else) does not break it further down the road.

Ok, I add a test case for embedded structure.

arl:
Is this feature going to get in master anytime soon? 
As said before, without the ability to flatten embedded stricts, it's tedious to re-use existing structures. 
Thanks

philhofer:
Committed in 21758201d17bec6caf42048839f88a434de734df

Issue #241, opened 2018-12-04
Not able to use msgp with go modules


Halfi:
Command `go get -u -t github.com/tinylib/msgp ` install msgp from master. It generates code with `msgp.WrapError`.
But when using msgp in app with go modules, latest version is v1.0.2 and there are no `msgp.WrapError`. To migrate to version v2 you should use `v2` postfix in import path, but I can't modify pregenerated code.
It was broken in commit 1360bdadb5d921b3394fa093ea4f2c9a43e523a9

philhofer:
The code doesn't officially support go modules (at the moment).

In either case, the tool and the library that comes with it are bundled together; you must vendor them using the same strategy. The fact that go modules don't recognize this coupling is, IMO, a deficiency in the design of the module system. In any case, maintaining a compatibility matrix of tool and library versions sounds like a huge pain, so I don't see this tool/library supporting that in the future unless someone is willing to sponsor the work. My guess is that the application in question can be fixed if you can encourage the go toolchain to pull the whole project as a dependency.

With that being said, I have been meaning to port this project to modules so that all the dependencies get pulled in correctly. (There's a 'v2' branch I've been patching on and off for a while now.)

TL;DR the library code and code generator are inseparable; the only stability guarantees I am providing are source-level backwards-compatibility and encoding backwards-compatibility. That is, for any existing application, you can upgrade to the new library _and code generator_, re-generate the code, and everything will continue to work as before.

Halfi:
Still I can use this library, but with one discomfort thing. I have to install msgp with some hack. Install with command `go get -u github.com/tinylib/msgp@v1.0.2`. But I can install it only if I have go.mod in the workdir. Looks like in 1.12 it have been fixed (but at now it still beta).

To use `v2` the auto generated code should use import like this `import "github.com/tinylib/msgp.v2"` but it impossible without fixes generator.

PR #252, opened 2019-05-11
v1: add go.mod and go.sum to track dependencies

urandom2:
While the status of v2 is outstanding, see #241, some customers depend
on the v1 line of releases, and it would be nice to track module
dependencies explicitly and prevent // indirect dependencies.

After merge can you release `v1.1.1`, so consumers can pickup the change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,9 @@
+module github.com/tinylib/msgp
+
+go 1.12
+
+require (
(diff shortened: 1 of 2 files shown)


thaJeztah:
Looks like this was addressed in https://github.com/tinylib/msgp/pull/281

PR #253, opened 2019-05-28
Type alias.

vrealzhou:
Added support for type alias. The type alias will be treat as same as source type during code generation.


vrealzhou:
Can't handle alias from other packages

Issue #255, opened 2019-06-27
Go vet errors on autogenerated tests for structs with sync.Mutex field


shawnps-sigsci:
Hi, I have a struct with a sync.Mutex field like so:

```
type Foo struct {
  lock sync.Mutex
}
```

and when `msgp` generates test code for this struct, it calls `t.Logf`:

```
v := Foo{}
var buf bytes.Buffer
msgp.Encode(&buf, &v) 
m := v.Msgsize()
if buf.Len() > m { 
        t.Logf("WARNING: Msgsize() for %v is inaccurate", v)
}  
```

But `go vet` gives us an error because of this:

`call of t.Logf copies lock value`

Could this be fixed in `testgen.go` by doing something like this instead?

```
template.Must(encodeTestTempl.Parse(`func TestEncodeDecode{{.TypeName}}(t *testing.T) {
v := {{.TypeName}}{}
var buf bytes.Buffer
msgp.Encode(&buf, &v) 

m := v.Msgsize()
if buf.Len() > m { 
        t.Log("WARNING: TestEncodeDecode{{.TypeName}} Msgsize() is inaccurate")
}   
```

Happy to make a pull request if you're on board with this proposal.

philhofer:
Yep, I'd be happy to accept that PR.

shawnps-sigsci:
Thank you @philhofer. I've created the PR here:

https://github.com/tinylib/msgp/pull/256

I've tested it out locally and it fixes the `go vet` copylocks error for me.

PR #256, opened 2019-06-27
gen: prevent generated tests from calling t.Logf with struct as argument, fixing go vet copylocks error when struct has sync.Mutex field - #255

shawnps-sigsci:
https://github.com/tinylib/msgp/issues/255

### DIFF
--- a/_generated/vet_copylocks.go
+++ b/_generated/vet_copylocks.go
@@ -1,0 +1,11 @@
+package _generated
+
+import "sync"
+
+//go:generate msgp
+//go:generate go vet
+
+type Foo struct {
+	I    struct{}
+	lock sync.Mutex
+}
--- a/gen/testgen.go
+++ b/gen/testgen.go
@@ -129,7 +129,7 @@ func BenchmarkUnmarshal{{.TypeName}}(b *testing.B) {
 
 	m := v.Msgsize()
 	if buf.Len() > m {
-		t.Logf("WARNING: Msgsize() for %v is inaccurate", v)
+		t.Log("WARNING: TestEncodeDecode{{.TypeName}} Msgsize() is inaccurate")
 	}
 
 	vn := {{.TypeName}}{}


philhofer:
Code change itself LGTM. Would love to have a test case as well (there are other script-driven test cases in the _generated directory you can take a peek at; I think you can just add the repro type definition and then stick `//go:generate go vet` in the file after the `//go:generate msgp` line).

shawnps-sigsci:
@philhofer unfortunately that surfaced some `go vet` errors in the actual generated code as well:

```
./vet_copylocks_gen.go:39:9: EncodeMsg passes lock by value: github.com/tinylib/msgp/_generated.Foo contains sync.Mutex
./vet_copylocks_gen.go:49:9: MarshalMsg passes lock by value: github.com/tinylib/msgp/_generated.Foo contains sync.Mutex
./vet_copylocks_gen.go:87:9: Msgsize passes lock by value: github.com/tinylib/msgp/_generated.Foo contains sync.Mutex
```

Do you know why it's generating some methods using `*Foo` and others with `Foo`?

```
➜  _generated git:(go-vet-fix-copylocks-testgen) ✗ grep Foo vet_copylocks_gen.go 
func (z *Foo) DecodeMsg(dc *msgp.Reader) (err error) {
func (z Foo) EncodeMsg(en *msgp.Writer) (err error) {
func (z Foo) MarshalMsg(b []byte) (o []byte, err error) {
func (z *Foo) UnmarshalMsg(bts []byte) (o []byte, err error) {
func (z Foo) Msgsize() (s int) {
```

shawnps-sigsci:
This is my `vet_copylocks.go`:

```
➜  _generated git:(go-vet-fix-copylocks-testgen) ✗ cat vet_copylocks.go 
package _generated

import "sync"

//go:generate msgp
//go:generate go vet

type Foo struct {
	lock sync.Mutex
}
```

shawnps-sigsci:
@philhofer so if I embed a struct inside `Foo` like this:

```
package _generated

import "sync"

//go:generate msgp
//go:generate go vet

type Inner struct {
	Name string
}

type Foo struct {
	I    Inner
	lock sync.Mutex
}
```

all is well.

If this is acceptable I'll update the PR.

philhofer:
There have been a bunch of requests in the past to pass stuff by value instead of by reference, and I (foolishly, perhaps) acquiesced.

For now, I think your workaround to prevent the code generator from passing your repro by value is fine. I'll have to fix the code generator to elide copies for structs that contain any of the `sync.*` types.

shawnps-sigsci:
@philhofer I see, thank you - how does the most recent commit look? Is that what you meant by adding a test script which calls `go vet` after generating?

philhofer:
Yep, looks good. Thanks!

Issue #263, opened 2019-10-01
omitempty support


bradleypeabody:
I want to check if support for "omitempty" can be reconsidered. I've looked through the earlier discussion on this, e.g. https://github.com/tinylib/msgp/issues/103#issuecomment-122414776

Specifically:

> For each of the five methods that we generate today, we would have to make substantial changes. Additionally, many optimizations (constant propagation and pre-serialization in particular) depend on knowing statically the number of elements to be encoded. For some users, the performance penalty may be acceptable, but we would have to make sure people understood the tradeoffs that they were making.

These points make sense. (Although I think it's only the methods related to writing, not reading, no?)  I think it would still make sense to implement as a feature. 

It seems to me the primary difference is whether one is **optimizing for size or encoding speed**.  In my situation, I have a complex series of structs with a lot of fields describing the state of an IoT device, but in many cases a lot of fields are empty.  These then end up msgpack'ed in a database field and read many times.  Compacting this information to reduce data size read/written to/from disk and transferred over the network is much more important for performance here than initial encoding speed.  The only immediate solution I see right now is to implement EncodeMsg by hand.

That said, how about the following for "omitempty" support:

* Existing generated code for EncodeMsg and MarshalMsg is exactly the same.
* Adding `msg:"myfield,omitempty"` on one or more struct fields triggers the additional logic of checking for empty field values before writing the map header.
* And of course the `if` statements that go around each field for empty check.
* Then this fact is just documented, e.g. "Using 'omitempty' on any field emits additional code to check for empty values and can reduce write performance in favor of producing smaller output.  (Performance implications of course depend on your use case, but see here for a benchmark ... ~X% is a sensible estimate.)".

Does that seem workable?  I don't really see a down side as all of the existing gains from what you have remain intact, it just allows this tradeoff if size reduction is more important.  If you can point me in the right direction of where to make these changes I'll see if I can get in there and make a prototype, get together a PR and we can discuss what would be needed to merge.

philhofer:
As long as you've got a sense of where the sharp edges are in this
feature, feel free to take a crack at it.

bradleypeabody:
[...]

PR #264, opened 2019-10-10
PR for omitempty support

bradleypeabody:
As discussed in #263, here is a PR for omitempty support.

Highlights:

- types moved to more appropriate places - bmask added a type in spec.go, the struct tag checking was added directly to Struct and StructField types, ZeroExpr and IfZeroExpr added to Elem
- as before, no change if not using "omitempty"
- no change to decoding behavior
- implementation and test stuff added for various cases like embedded structs

Let me know if anything else is needed on this, but I believe this hits all of the important points.

### DIFF
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -1,8 +1,10 @@
 package gen
 
 import (
+	"bytes"
 	"fmt"
 	"io"
+	"math"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


bradleypeabody:
Here is my suggestion for documentation:

### omitempty support

Tagging a field on a struct with `msg:"fieldname,omitempty"` will cause the code generator to emit additional code to check if the field is empty (Go zero value) before writing it.  The behavior of this option generally attempts to emulate that of the [encoding/json](https://golang.org/pkg/encoding/json/) package.  Example:

```
type Example struct {
    A string `msg:"a,omitempty"`
    B int    `msg:"b,omitempty"`
    C string `msg:"c"`
}
```

In this example, A will be written if not `""`, B will be written if not `0` and C will always be written.

The omitempty option is ignored if used on a field where it is not supported.

The check for zero/empty is defined as `false`, `0`, `""`, `complex(0,0)`, `nil`, `time.Time{}` or `StructTypeName{}`, according to type.  Arrays and structs with no type name are not supported.  Note that, unlike encoding/json, named structs are supported.

It should be understood there are performance tradeoffs when using the omitempty option.  This is because the generated code must check fields to see if they are empty, write the map header with the correct length, and then do a separate pass to write each non-empty field. (Structs with no omitempty fields do not cause this additional code to be emitted.) Feel free run the benchmarks in _generated/omitempty_test.go for an example of the speed difference. A 10-15% speed decrease for the case of all fields being not-empty is a reasonable estimate.  For the case of the majority of the fields being empty, it may perform significantly faster since field names and zero-length contents need not be written.

Decoding is not affected by this option.  Fields omitted from the encode with this option are simply not touched during decode.  Note that this means if you have existing code that re-uses the same struct instance, decoding into it multiple times, you'll need to zero it out (assign `StructName{}` to it) before decoding in order to get correct results using omitempty.

gen/spec.go
```diff
@@ -432,3 +434,91 @@ func (p *printer) ok() bool { return p.err == nil }
...
+		return "uint16"
+	}
+	if b.bitlen <= 32 {
+		return "uint32"
+	}
+	if b.bitlen <= 64 {
+		return "uint64"
+	}
+
+	return fmt.Sprintf("[%d]uint64", (b.bitlen>>6)+1) // (bitlen/64)+1
```
philhofer: the expression you want is `(b.bitlen + 64 - 1) / 64` so that, for example, 128 yields `[2]uint64` rather than `[3]uint64` as it would now

(don't worry about manually strength-reducing the integer math; the compiler will do this for you)

gen/spec.go
```diff
@@ -432,3 +434,91 @@ func (p *printer) ok() bool { return p.err == nil }
...
+}
+
+// readExpr returns the expression to read from a position in the bitmask.
+// Compare ==0 for false or !=0 for true.
+func (b *bmask) readExpr(bitoffset int) string {
```
philhofer: I'd include a sanity check here that `bitoffset <= b.bitlen`

gen/spec.go
```diff
@@ -432,3 +434,91 @@ func (p *printer) ok() bool { return p.err == nil }
 func tobaseConvert(b *BaseElem) string {
 	return b.ToBase() + "(" + b.Varname() + ")"
 }
+
+func (p *printer) varMapHeader(pre, post string, sizeVarname string, maxSize int) {
+
+	if maxSize <= 15 {
+		p.printf("\n%s 0x80 | (uint8(%s)&0x0F) %s", pre, sizeVarname, post)
+	} else if maxSize <= math.MaxUint16 {
```
philhofer: I think when `maxSize` is above 15, you should just fall back to emitting a call to `msgp.AppendMapHeader` (or its analog for encoding). The functional call overhead will be amortized over the large number of struct fields.

bradleypeabody: Agreed - and the compiler might very well inline msgp.AppendMapHeader anyway.

gen/spec.go
bradleypeabody: Good catch, fixed.

gen/spec.go
bradleypeabody: Done

gen/spec.go
bradleypeabody: (I checked and the compiler does not appear to be inlining msgp.AppendMapHeader, at least in the case I looked at, but I agree the unnecessary bit operations here are more likely to result in an obscure bug than a measurable performance improvement).

Fixed.

bradleypeabody:
The points brought up earlier today are all fixed.

bradleypeabody:
Please see notes at #263

bradleypeabody:
Thanks @philhofer ! It was great working with you on this.

philhofer:
Cheers. Thanks for the PR.

Issue #269, opened 2020-03-26
errWrapped should have an Unwrap method


bradleypeabody:
The latest changes to https://golang.org/pkg/errors/ provide for an error wrapping mechanism in the Go standard library.  Things should work as expected with `errors.Is()` and other such methods if an `Unwrap() error` method is added to errWrap (see https://github.com/tinylib/msgp/blob/master/msgp/errors.go#L106 ).

Let me know if you agree and I can make a PR for it.

philhofer:
Yes! Sounds good.

PR #270, opened 2020-03-26
added Unwrap method to errWrapped plus tests; switched travis to go 1.14

bradleypeabody:
Here's the PR as mentioned in https://github.com/tinylib/msgp/issues/269

Since the new error handling requires at least Go 1.13, I went ahead and updated the travis config to the latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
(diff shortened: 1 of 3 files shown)


philhofer:
Looks good. Thanks!

PR #276, opened 2020-07-30
Add Reader/Writer constructors with custom buffer

VirrageS:
This patch adds simple new functions to create `Reader`/`Writer` with custom buffers. It is helpful in times when buffers are recycled with `sync.Pool` this way we can reuse buffer rather than having separate `sync.Pool` for `msgp.Reader`/`msgp.Writer`.

This change doesn't change any existing logic.

~~Requires https://github.com/philhofer/fwd/pull/19 to be merged.~~

Once this MR is accepted and merged I would really appreciate to bump the version (it would help to update the dependency in our project).

### DIFF
--- a/msgp/write.go
+++ b/msgp/write.go
@@ -10,6 +10,11 @@ import (
 	"time"
 )
 
+const (
+	// min buffer size for the writer
+	minWriterSize = 18
+)
+
 // Sizer is an interface implemented
 // by types that can estimate their
 // size when MessagePack encoded.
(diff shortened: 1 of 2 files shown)


msgp/write.go
```diff
@@ -120,16 +125,25 @@ func NewWriter(w io.Writer) *Writer {
...
+		sz = minWriterSize
 	}
+	buf := make([]byte, sz)
+	return NewWriterBuf(w, buf)
+}
 
+// NewWriterBuf returns a writer with a provided buffer.
+// It panics if the buffer size is smaller than 'minWriterSize'.
+func NewWriterBuf(w io.Writer, buf []byte) *Writer {
+	if len(buf) < minWriterSize {
```
philhofer, line 141: In principle you can stretch `buf` to `buf[:cap(buf)]` here.

But I think it would be fine to reallocate the buffer if its capacity is actually under 18 bytes, provided that the API and docs convey that information appropriately.

VirrageS: Updated

VirrageS:
Any update on this one?

VirrageS:
Bump

VirrageS:
Bump. @philhofer could you look into this?

philhofer (approved):

PR #278, opened 2020-08-31
Always create slices, even 0 length

### LINKED ISSUE #247: Decode replaces empty slices with nil
@philhofer @erikdubbelboer https://github.com/tinylib/msgp/issues/219 exists in the case of empty slices, where unmarshalling 0-len slices turn them into `nil`s. Could this behaviour be fixed to be consistent with that of maps?

```go
package main

import (
	"fmt"
)

//go:generate msgp

type Test struct {
	Slice []string          `msg:"slice`
	Map   map[string]string `msg:"map"`
}

var b1, b2 []byte

func main() {
	// Slice and Maps are initialized (non-nil) but have empty values.
	pre := Test{Slice: []string{}, Map: map[string]string{}}
	fmt.Println("pre |", ".Slice == nil ?", pre.Slice == nil, "| .Map == nil ?", pre.Map == nil)

	var (
		b, _ = pre.MarshalMsg(nil)
		post = Test{}
	)
	post.UnmarshalMsg(b)

	// .Slice and has become nil.
	fmt.Println("post |", ".Slice == nil ?", post.Slice == nil, "| .Map == nil ?", post.Map == nil)
}
```

Output:
```
pre | .Slice == nil ? false | .Map == nil ? false
post | .Slice == nil ? true | .Map == nil ? false
```

erikdubbelboer:
This change makes it so that when a 0 length slice is present in the
msgp data a 0 length slice will be allocated in Go, instead of nil.

This means msgp data generated using this library will always result in
a slice being allocated as this library doesn't contain any conditional
code to not emit the slice header if a slice is nil. Other libraries
that do omit the slice header will still result in a nil slice if the
slice is completely missing from the msgp data.

Fixes #247

@philhofer you might also want to think about changing the marshal code to not at all emit the slice/map headers if the slice/map is `nil`. This will result in less data and would allow users to differentiate between `nil` and empty.
Both this change and that suggested change are backwards incompatible if users depend on a certain `nil`/empty behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 # normal `go install`.
 
 # generated integration test files
-GGEN = ./_generated/generated.go ./_generated/generated_test.go
+GGEN = ./_generated/generated.go ./_generated/generated_test.go ./_generated/*_gen.go ./_generated/*_gen_test.go
 # generated unit test files
 MGEN = ./msgp/defgen_test.go
 
--- a/_generated/gen_test.go
+++ b/_generated/gen_test.go
@@ -97,9 +97,12 @@ func Test1EncodeDecode(t *testing.T) {
 			ValueA: "here's the first inner value",
 			ValueB: []byte("here's the second inner value"),
 		},
-		Child:    nil,
-		Time:     time.Now(),
-		Appended: msgp.Raw([]byte{}), // 'nil'
+		Child:     nil,
+		Time:      time.Now(),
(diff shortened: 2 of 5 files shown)


philhofer:
I'm going to have to think for a bit on the compatibility implications of this change.
Up to this point there hasn't been a good reason to start a v2.0 series, but at this point it might be worth it so that we can evaluate a few small breaking changes.

PR #279, opened 2020-09-17
CopyToJSON: fix bitSize for floats

hectorj:
It was reversed for `float32` and `float64`.

Added a test case too


philhofer:
Nice. Thanks for the bugfix!

PR #281, opened 2020-10-22
travis: use go1.15.x; add go.mod

philhofer:


.travis.yml
```diff
@@ -3,22 +3,15 @@ language: go
...
-  - GIMME_ARCH=amd64
-  - GIMME_ARCH=386
-  - GIMME_ARCH=ppc64le
-jobs: 
-  exclude:
-    - arch: ppc64le
-      env:  GIMME_ARCH=amd64
-    - arch: ppc64le
-      env:  GIMME_ARCH=386
+jobs:
```
klauspost, line 12:
```suggestion
jobs:
  allow_failures:
    - go: 'tip'
```

This should would build even if tip is broken and should give time to fix issues.

PR #288, opened 2021-04-21
Feature: New msgp option to recursively compile all Go files in a directory

rafaeljusto:
Allow providing a directory to recursively compile all Go files. When the new `dir` argument is provided the arguments `file` nd `out` will be ignored.

This is useful to have only one `go generate` for all your project in the main file. For example:

```go
package main

//go:generate go get -u github.com/tinylib/msgp
//go:generate msgp -io=false -tests=false -dir=../../internal/entities

func main() {
  // ...
}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -117,18 +117,20 @@ func compileDir(dir string, mode gen.Method) {
 	}
 	defer d.Close()
 
-	fis, err := d.Readdir(-1)
+	fid, err := d.ReadDir(-1)
 	if err != nil {
 		exitln("Cannot read files in " + dir + ": " + err.Error())
 	}
 
 	var names []string
-	for _, fi = range fis {
-		name := fi.Name()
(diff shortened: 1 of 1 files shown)


rafaeljusto:
The downside would be trying to compile Go files that we know don't support `msgp`. 🤔 

For example:
```go
type Example struct {
  Field SomeType
}

type SomeType interface {
  DoIt()
}
```

That would result:
```go
func (z *Example) MarshalMsg(b []byte) (o []byte, err error) {
  // ...
  o, err = z.Field.MarshalMsg(o)
  // ...
}
```

And `MarshalMsg` is not part of the interface.

rafaeljusto:
> ./main.go:120:15: d.ReadDir undefined (type *os.File has no field or method ReadDir, but does have Readdir)

Sorry, I don't know we were compiling with Go 1.14. I will go back to use `Readdir`.

rafaeljusto:
Never mind, you can already use a directory in the file parameter 🤦 
> msgp -io=false -tests=false -file=../../internal/entities

Closing.

PR #289, opened 2021-05-21
Add missing bin header functions

klauspost:
There is no bytes equivalent of `(m *Reader) ReadBytesHeader()` and `(mw *Writer) WriteBytesHeader`.

Add `ReadBytesHeader(b []byte)` and `AppendBytesHeader(b []byte, sz uint32)`.


philhofer:
Looks good. Thanks.

Issue #294, opened 2021-07-21
TinyGo support


bradleypeabody:
The msgp package currently does not compile under [TinyGo](https://tinygo.org).   Messagepack support on microcontrollers and in WASM I think is super useful and it would be great if msgp+tinygo worked.

The required fixes seem quite minimal - so far from my initial testing, just build tags:

```
diff --git a/msgp/advise_linux.go b/msgp/advise_linux.go
index 6c6bb37..7b6049f 100644
--- a/msgp/advise_linux.go
+++ b/msgp/advise_linux.go
@@ -1,4 +1,4 @@
-// +build linux,!appengine
+// +build linux,!appengine,!tinygo

 package msgp

diff --git a/msgp/file.go b/msgp/file.go
index 8e7370e..54586ba 100644
--- a/msgp/file.go
+++ b/msgp/file.go
@@ -1,5 +1,5 @@
 // +build linux darwin dragonfly freebsd netbsd openbsd
-// +build !appengine
+// +build !appengine,!tinygo

 package msgp

diff --git a/msgp/file_port.go b/msgp/file_port.go
index 6e654db..3e33b1f 100644
--- a/msgp/file_port.go
+++ b/msgp/file_port.go
@@ -1,4 +1,4 @@
-// +build windows appengine
+// +build windows appengine tinygo

 package msgp
```

And a couple for the `fwd` package as well:
```
diff --git a/writer_appengine.go b/writer_appengine.go
index e367f39..fa6e90a 100644
--- a/writer_appengine.go
+++ b/writer_appengine.go
@@ -1,4 +1,4 @@
-// +build appengine
+// +build appengine tinygo

 package fwd

diff --git a/writer_unsafe.go b/writer_unsafe.go
index a0bf453..1701572 100644
--- a/writer_unsafe.go
+++ b/writer_unsafe.go
@@ -1,4 +1,4 @@
-// +build !appengine
+// +build !appengine,!tinygo

 package fwd
```

If I do PRs to fix these things, is this something you would be able to merge?  I realize it may not be possible to get 100% of the functionality under TinyGo, but it seems like the vast majority of this can be made to work easily with just some build tags.  If you're good with the idea, I can do the work and send PRs.

philhofer:
Yes, those changes seem perfectly fine. I'd be happy merging them.
(But even happier if we had some way to make sure we were covering these builds in our continuous integration so that we don't regress on this support once we have it.)

bradleypeabody:
Awesome.

[...]

PR #295, opened 2021-07-24
TinyGo support

bradleypeabody:
So this (https://github.com/tinylib/msgp/issues/294) turned out to be more involved than I originally thought, but the changes are actually fairly simple:

* **go.mod updated** to use latest `fwd`
* `file*` and `advise*` were **just build tag changes**
* **elsize.go needed an alternate implementation** for TinyGo - the `sizes` lookup table makes sense on regular machines but on a microcontroller it is better to move that into code in the form of a switch statement and save the 1k of RAM.  References in other files to `sizes[x]` were converted to `getBytespec(x)`, which I verified with `go test -gcflags="-m=2"` are inlined by the regular Go compiler, e.g. `./read_bytes.go:1208:21: inlining call to getBytespec func(byte) bytespec { return sizes[v] }`
* **References to `fmt` package removed for TinyGo** compilation path.  In some cases I replaced with a call to `strconv` that does the same thing, and a couple others (see `ctxString`) I kept the original implementation and made a non-fmt version for TinyGo.  The `fmt` package adds about 30k to the output code size - in one example that's 70% of the overall program size.
* I checked a few other things like the `init()` methods in `json*.go` and the various dynamic stuff in the`WrapError` method to see if there were any size optimization opportunities there and everything turned out to be either already optimized away by the compiler or only have a negligible size difference, so I didn't take it any further.  Also, the buffers I was concerned about in `fwd` I don't think are relevant anyway - if someone is writing code where an extra couple of kb of memory matters, they should almost certainly not be using a Reader/Writer abstraction, but instead stick with Append..., MarshalMsg, UnmarshalMsg, etc.  I think this can just be mentioned in the wiki somewhere.

And the rest is the test suite (`tinygotest` dir).  Each subdir is a program and there are regular Go tests that drive building via `tinygo` for various target outputs and also build and run the executable on the local platform (tested on Mac, should work on Linux, won't work on Windows) and verify it exits cleanly.  This run step turned out to be needed as there was a subtle panic in the beginning that this helped catch and fix.   There is also some logging that shows the output file sizes so one can get a feel for how big the resulting binaries are on MCUs.   ARM and WASM outputs work without issue, but I ran into some problems with other architectures like AVR and ESP32/xtensa.  If I find the time I'll see if I can fiddle with it more but for now I think what's here is still very useful for a lot of environments (including very popular ones like ARM-based Arduinos).

I also updated `.travis.yml` to [install TinyGo](https://tinygo.org/getting-started/install/linux/), we'll see if it works.  And I added Go 1.16.x while I was in there.

### DIFF
--- a/msgp/elsize_tinygo.go
+++ b/msgp/elsize_tinygo.go
@@ -1,0 +1,12 @@
+// +build tinygo
+
+package msgp
+
+// for tinygo, getBytespec just calls calcBytespec
+// a simple/slow function with a switch statement -
+// doesn't require any heap alloc, moves the space
+// requirements into code instad of ram
+
+func getBytespec(v byte) bytespec {
+	return calcBytespec(v)
+}
--- a/msgp/errors_default.go
+++ b/msgp/errors_default.go
@@ -1,0 +1,24 @@
+// +build !tinygo
+
+package msgp
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// ctxString converts the incoming interface{} slice into a single string.
+func ctxString(ctx []interface{}) string {
+	out := ""
+	for idx, cv := range ctx {
+		if idx > 0 {
+			out += "/"
+		}
+		out += fmt.Sprintf("%v", cv)
+	}
+	return out
+}
+
+func quoteStr(s string) string {
+	return strconv.Quote(s)
+}
--- a/msgp/errors_tinygo.go
+++ b/msgp/errors_tinygo.go
@@ -1,0 +1,43 @@
+// +build tinygo
+
+package msgp
+
+import (
+	"reflect"
+)
+
+// ctxString converts the incoming interface{} slice into a single string,
+// without using fmt under tinygo
+func ctxString(ctx []interface{}) string {
+	out := ""
+	for idx, cv := range ctx {
+		if idx > 0 {
+			out += "/"
+		}
+		out += ifToStr(cv)
+	}
+	return out
+}
+
+type stringer interface {
+	String() string
+}
+
+func ifToStr(i interface{}) string {
+
+	switch v := i.(type) {
+	case stringer:
+		return v.String()
+	case error:
+		return v.Error()
+	case string:
+		return v
+	default:
+		return reflect.ValueOf(i).String()
+	}
+
+}
+
+func quoteStr(s string) string {
+	return simpleQuoteStr(s)
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: bionic
+
 language: go
 
 arch:
@@ -5,8 +7,8 @@ arch:
   - ppc64le
   - arm64
 go:
-  - 1.14.x
   - 1.15.x
+  - 1.16.x
   - tip
 
 jobs:
@@ -17,3 +19,4 @@ jobs:
       env: GIMME_ARCH=386
 
 script: "make travis"
+
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ all: install $(GGEN) $(MGEN)
 
 # travis CI enters here
 travis:
+	arch
+	if [ `arch` == 'x86_64' ]; then sudo apt update; sudo apt install build-essential; wget https://github.com/tinygo-org/tinygo/releases/download/v0.18.0/tinygo_0.18.0_amd64.deb; sudo dpkg -i tinygo_0.18.0_amd64.deb; export PATH=$PATH:/usr/local/tinygo/bin; fi
 	go get -d -t ./...
 	go build -o "$${GOPATH%%:*}/bin/msgp" .
 	go generate ./msgp
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,6 @@ module github.com/tinylib/msgp
 go 1.14
 
 require (
-	github.com/philhofer/fwd v1.1.1
+	github.com/philhofer/fwd v1.1.2-0.20210722190033-5c56ac6d0bb9
 	golang.org/x/tools v0.0.0-20201022035929-9cf592e881e9
 )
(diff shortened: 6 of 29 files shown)


bradleypeabody:
And, sorry about the mess on these last commits - I can squash/rebase if needed.    It took a bit to get the Travis build working, but it's looking good now: https://www.travis-ci.com/github/bradleypeabody/msgp

Changes were:
* TinyGo tests run on amd64 only (to avoid having to install `tinygo` on arm, etc.)
* Makefile now checks for arch and properly installs TinyGo .deb package
* had to upgrade to Ubuntu 18.04 (dist: bionic) due to https://github.com/tinygo-org/tinygo/issues/1064
* `go generate` on tinygotest folder, otherwise go test ./... fails on other arches due to missing MarshalMsg, etc. even though no test files
* I removed Go 1.14.x so it's just the most recent two versions 1.15, 1.16 - in order to save travis build minutes

msgp/elsize_tinygo.go
```diff
@@ -0,0 +1,129 @@
...
+package msgp
+
+// for tinygo, we use a simple/slow function with a switch
+// statement - doesn't require any heap alloc, moves the space
+// requirements into code instad of ram
+
+func getBytespec(v byte) bytespec {
+
+	// put this all in a big switch and let the compiler figure it out
+	switch {
```
philhofer, line 12: For the switch here, let's do `switch v { case mnil: ... }` and then have a second switch for the cases that don't match constants.

msgp/elsize_tinygo.go
```diff
@@ -0,0 +1,129 @@
...
+	case v >= mfixmap && v < 0x90:
+		return bytespec{size: 1, extra: varmode(2 * rfixmap(v)), typ: MapType}
+
+	// fixarray
+	case v >= mfixarray && v < 0xa0:
+		return bytespec{size: 1, extra: varmode(rfixarray(v)), typ: ArrayType}
+
+	}
+
+	panic(fatal)
```
philhofer, line 99: I can't tell if this case is reachable from the entry of this function, but we should take care that it isn't. Otherwise malformed input will lead to a panic here.

Might make sense to order these comparisons so that they are sequential (and therefore the last case of `uint16(v) < 0x100` is trivially true.

msgp/errors_tinygo.go
```diff
@@ -0,0 +1,43 @@
...
+	case string:
+		return v
+	default:
+		return reflect.ValueOf(i).String()
+	}
+
+}
+
+func quoteStr(s string) string {
+	return `"` + s + `"`
```
philhofer, line 42: This doesn't do quite the same thing as `%q` for control characters. Since `strconv` is already a dependency of the `tinygo` build, can we use `strconv.Quote()` here?

msgp/errors_default.go
```diff
@@ -0,0 +1,23 @@
...
+	for idx, cv := range ctx {
+		if idx > 0 {
+			out += "/"
+		}
+		out += fmt.Sprintf("%v", cv)
+	}
+	return out
+}
+
+func quoteStr(s string) string {
```
philhofer, line 21: This can just be `strconv.Quote()`

philhofer (changes requested):
Thanks for your hard work on this!

It would be great if you could squash away the travis debugging commits, and any other commits that don't pass tests.

Also, I was curious if it's possible to put the `tinygo` tests in `testdata/` or another directory that the regular `go` tool would ignore so that the code isn't built as part of a `go install ./...` or equivalent. (Another technique I have seen for this is sticking `//+build ignore` at the top of the files so that they are only built with an explicit `go run` or `go build`, but otherwise ignored.)

bradleypeabody:
Thanks for the feedback!  Made the following changes:

* I moved the TinyGo test programs to a `testdata` folder - this works great
* elsize/bytespec switch cleaned up - you're right about that panic - it was being called for the case of the unused byte - 0xC1.  That's fixed now. More data below.
* quoteStr reworked (see below)
* Everything squashed into one commit

---

Some more detail:

**On the elsize changes** - I refactored this a bit to move all of the bytespec logic into one function - `calcBytespec`, and then in TinyGo `getBytespec` just calls `calcBytespec`, but for non-TinyGo `init` uses `calcBytespec` to initialize the `sizes` lookup table.  This approach has the same result as the previous changes for both TinyGo and non-TinyGo environments, but moves the logic into one function so it doesn't diverge.  I then added a test case with the old code as proof that this change gives the exact same result as before.  Previously I was trying to minimize change to the existing elsize.go file, but I think this approach is the best balance of the various factors.

**For quoteStr,** the issue is that calling `strconv.Quote` results in a code path that hits the unicode lookup tables to determine which UTF-8 characters are printable as part of the escaping logic.  Since quoting is only used for error reporting, it seems overkill to bring this code in just to handle escaping field names in error messages.  So what I did is make a trimmed down function that follows the same escaping rules for ASCII, but any bytes > 127 are just escaped with `\x`.  This is easy to implement and I think achieves the desired result.  I measured the TinyGo output code sizes to help justify this approach:

* **Naive implementation**
```
`"` + s + `"`
```
tinygo_test.go:246: size report - arduino-nano33.bin        13144 bytes

* **Using stdlib**
```
strconv.Quote("a")
```
tinygo_test.go:246: size report - arduino-nano33.bin        18164 bytes

* **Custom function**
```// TinyGo-specific simplified quoteStr implementation
tinygo_test.go:246: size report - arduino-nano33.bin        13612 bytes

```

msgp/elsize_tinygo.go
bradleypeabody: fixed

msgp/elsize_tinygo.go
bradleypeabody: agreed - fixed in the refactor

msgp/errors_tinygo.go
bradleypeabody: see notes on strconv.Quote code size

msgp/errors_default.go
bradleypeabody: fixed

philhofer (approved):

philhofer:
Thanks!

bradleypeabody:
No problem!  FYI I went ahead and added a mention of this in the docs here: https://github.com/tinylib/msgp/wiki/Tips-and-Tricks#tinygo-support

PR #299, opened 2021-10-01
Bytes conversion issue

jhwz:
Allows the 'using' method to be used from a shim


klauspost:
@jhwz I would like to add this, but I don't know what I'm looking at.

Could you

A) Describe what this does.
B) Add a test for the functionality?

jhwz:
Goodness, I can't really remember, it's been a while. I'll try look into this later this week - I appreciate the work you guys do!

klauspost:
If this is still an issue, please open a new PR, which includes a reproducer test.

PR #301, opened 2021-10-17
fix time.Time to msgpack.spec

piyongcai:
0. fix time.Time to msgpack.spec
   a. reade_bytes.go  -> ReadTimeBytes
   b. write_bytes.go  -> AppendTime
   c. simple_marshal
   d. simple_unmarshal
1. change TimeExtension from 5 -> 255(-1)
2.    var extensionReg = make(map[int8]func() Extension)
   -> var extensionReg = make(map[uint8]func() Extension)
3. type ExtensionTypeError struct {
	Got  int8  // -> uint8
	Want int8  // -> uint8
    }
4. type Extension interface {
	ExtensionType() int8  // -> uint8
	Len() int
	MarshalBinaryTo([]byte) error
	UnmarshalBinary([]byte) error
  }
5. ype RawExtension struct {
	Data []byte
	Type int8   //-> uint8
  }

AND Other related changes


klauspost:
(Code review appears broken)

You can't just dump type 5 and break everything already using this package. You should have proper fallback.

Regarding types, spec has this as a signed int8. So I don't see why you are changing it to uint8.

PR #302, opened 2021-10-26
Add allownil tag

klauspost:
Allows to specify `msgp:"name,allownil"` which will add nil entries for slices and maps. This also makes behavior more close to `encoding/json`.

This will allow symmetric roundtrips with tuples and regular objects without `omitempty` which will completely omit the key.

When specified alongside `omitempty` the struct will be omitted, so that takes precedence. To unmarshal maps and slices that can be nil, the `allownil` must be specified.

Unless specified no generated code changes.

Related #278

### DIFF
--- a/_generated/allownil.go
+++ b/_generated/allownil.go
@@ -1,0 +1,230 @@
+package _generated
+
+import "time"
+
+//go:generate msgp
+
+type AllowNil0 struct {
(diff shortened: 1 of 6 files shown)


lxyshuai:
what about `[]byte`？allownil in `[]byte` is not useful

PR #303, opened 2021-11-26
time.Time follow the MsgPack spec

piyongcai:
Add MarshalMsg, UnmarshalMsg,EncodeMsg,DecodeMsg,CopyToJSON,UnmarshalAsJSON to follow MsgPack spec.
the tests were also added.

### DIFF
--- a/msgp/extension.go
+++ b/msgp/extension.go
@@ -14,11 +14,11 @@ const (
 	Complex128Extension = 4
 
 	// TimeExtension is the extension number used for time.Time
-	TimeExtension = 5
+	TimeExtension = 255
 )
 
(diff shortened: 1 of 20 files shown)


msgp/extension.go
```diff
@@ -76,7 +76,7 @@ type Extension interface {
 	// type of the extension. (Types <0 are
 	// officially reserved by the MessagePack
 	// specifications.)
-	ExtensionType() int8
+	ExtensionType() uint8
```
klauspost, line 79: 1) You cannot change an exported interface.
2) Extension types are defined as signed in spec. Why this change at all?

PR #304, opened 2021-11-26
Make array on decode if it's not supposed to be nil

makarchuk:
Addresses https://github.com/tinylib/msgp/pull/278. #302  fixed issue only partially. It serialized non-nil empty `slice` as it should, but on decode it deserialized as `nil` anyway. 

This PR addresses this "issue".

I'm open to any suggestions on code improvement for this one

### DIFF
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -376,6 +376,7 @@ func (p *printer) wrapErrCheck(ctx string) {
 }
 
 func (p *printer) resizeSlice(size string, s *Slice) {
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM

makarchuk:
Can I help out somehow to speed up merge of this one? 
It's quite important for me

klauspost:
I suppose a few tests would help.

klauspost:
@makarchuk I have added #336 

When looking at this, I would expect the additional allocations to be a significant downside for current users. Instead I've added it to `allownil`. Since these members can explicitly state if they want to be empty slice or nil, it makes sense there.

That would make `allownil` members fully roundtrip safe in Go.

PR #310, opened 2022-06-16
added blank identifier to mask to prevent edge case

Alexander-Leon:
There's an edge case where if all the fields within a struct provide their own marshal/encode methods the mask doesn't end up being used at all which then causes the compiler "unused variable" flags to prevent compilation. This doesn't change the base operation of the functionality, but avoids needing to go manually adjust the auto generated code when this edge case comes up.


philhofer:
Thanks!

PR #311, opened 2022-07-16
Support time duration

arl:
Hi, we've been needing to add msgpack to a large codebase using many POD structures, some of them had `time.Duration` fields.

We haven't found a way to directly use `time.Duration` and it seemed like a good addition.
We're just sending the int64 value and converting at the edge, we don't decode the string version (eg `3m30s`).

Is this something you'd be interested in being merged upstream? If so, let me know what/if should I add anything to the current pull-request.

By the way, thanks for a very well-thought, easy to use and performant library.

### DIFF
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -88,9 +88,10 @@ const (
 	Int32
 	Int64
 	Bool
-	Intf // interface{}
-	Time // time.Time
-	Ext  // extension
+	Intf     // interface{}
+	Time     // time.Time
+	Duration // time.Duration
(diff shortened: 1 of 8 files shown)


klauspost:
I honestly can't think of any reasons not to add this.

Representing a `time.Duration` as an int64 of course looses the type but makes it much cleaner than an extension (and more compact as a bonus).

klauspost (approved):
LGTM

philhofer (approved):

PR #334, opened 2023-10-27
omitzero implementation

bradleypeabody:
Here is a PR for `omitzero` support, following the discussion at https://github.com/tinylib/msgp/pull/326.

Highlights are:

* `omitzero` is now separate from `omitempty`, the former emits a function call to IsZero(), and the latter checks against the Go zero value.  They both use the same bitmask and other related logic, only the actual check to determine if the field should be omitted is affected.  I also adjusted a few comments to reflect these changes, nothing major.
* Overall this change of making these two separate features simplified the PR significantly, which is awesome.
* The the name of Elem.IfZeroExpr and the fact that it is used by the `omitempty` feature and not by the `omitzero` feature is unfortunate, but I don't think a deal-breaker.  Probably better to just leave it and keep it simple, rather than trying to refactor the name of IfZeroExpr to fix this minor naming oddity.
* Since omitzero is only used for custom types, it reduced what I had in there in the way of tests - I basicaly removed all of the stuff related to primitive types since it's incompatible with IsZero().  As requested, I also added the check to ensure tuples are unaffected.

Let me know if anything else is needed for this, but hopefully that addresses all the concerns.

### DIFF
--- a/_generated/omitzero.go
+++ b/_generated/omitzero.go
@@ -1,0 +1,86 @@
+package _generated
+
+import "time"
+
+//go:generate msgp
+
+// check some specific cases for omitzero
+
+type OmitZero0 struct {
+	AStruct    OmitZeroA       `msg:"astruct,omitempty"`    // leave this one omitempty
+	BStruct    OmitZeroA       `msg:"bstruct,omitzero"`     // and compare to this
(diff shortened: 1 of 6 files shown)


klauspost (approved):
LGTM

PR #336, opened 2023-10-27
allownil: Allocate 0 length slices

klauspost:
When `allownil` is enabled, always allocate zero length slices.

This ensures roundtrips with 0-length slices are not returned as nil.

Replaces #304

Adds tests. Bonus: Don't shell out to test issue 94.


klauspost:
(tests require #335 )

klauspost:
With no objections I will humbly merge this.

PR #340, opened 2024-02-23
support for any maps in msgp.AppendIntf like in (*Writer).WriteIntf

eliottness:
Hello :wave:

Previously when using `AppendIntf` with types like `map[string][]any` the function would throw an `ErrUnsupportedType` error.

So I added support for all values in maps using the `reflect` package just like it was already done in `(*Writer).WriteIntf` (by basically copying the code from `(*Writer).writeMap`).

I did not see any tests for this function so I was not comfortable to start writing a new testsuite. Please let me know if you want me to test my code.

### DIFF
--- a/msgp/write_bytes.go
+++ b/msgp/write_bytes.go
@@ -1,6 +1,7 @@
 package msgp
 
 import (
+	"errors"
 	"math"
 	"reflect"
 	"time"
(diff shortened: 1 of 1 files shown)


klauspost:
Would it be possible to add a few tests for this?

klauspost (approved):
Looks harmless. I can live without the tests, though it would be great if you sent a PR with them.

eliottness:
@klauspost You merged while I added the tests :sob:
Opening a new PR

PR #341, opened 2024-02-26
test AppendIntf against ReadIntfBytes 

eliottness:
Adding new tests for `AppendIntf` and `ReadIntfBytes`. Follow up of #340


klauspost (approved):
Awesome! LGTM

PR #351, opened 2024-07-01
Add Go v1.22.x to CI workflow

VirrageS:
This adds `1.22.x` Go version to CI workflow so we also test against this version to make sure everything works fine.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -21,7 +21,7 @@ jobs:
   linters:
     strategy:
       matrix:
-        go-version: [1.21.x]
+        go-version: [1.22.x]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
@@ -35,5 +35,5 @@ jobs:
       - name: lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.55.1
+          version: v1.57.2
           args: --print-resources-usage --timeout=10m --verbose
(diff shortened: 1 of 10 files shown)


klauspost:
I think it is ok to remove 1.18 support. It is far beyond EOL.

.github/workflows/validate.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   linters:
     strategy:
       matrix:
-        go-version: [1.21.x]
+        go-version: [1.21.x, 1.22.x]
```
klauspost, line 24:
```suggestion
        go-version: [1.22.x]
```
Just latest should be fine for this.

VirrageS:
@klauspost It is your call, I can keep v1.18 for now.

klauspost:
Remove it in the PR. Having 4 versions back should be more than enough.

VirrageS:
From what I see the https://github.com/philhofer/fwd/commit/fbbf4953d98612dac320a9051da34e3e761d1df7 requires at least 1.20. And from what I see 1.22 build requires that we use this new `unsafe.Slice`. So I think we would need to drop 1.19 too :(

klauspost:
I'm not going to object to that. Personally I support last 3 versions, which is already one more than Go, but I don't know if Phil has an opinion on this.

I suggest you add that to this PR, we can adjust if needed.

klauspost (approved):
LGTM

philhofer:
Dropping support for EOL'd go versions is fine to me. LGTM.

PR #353, opened 2024-07-02
Added time.duration case to AppendIntf

RiaanLaubscher:
Adding supporting for encoding a `time.Duration` type.


klauspost (approved):
LGTM

PR #363, opened 2024-09-18
Create 0 size byte slices on allownil []byte elements

### LINKED ISSUE #362: Allownil in []byte is useless

klauspost:
When `allownil` is specified on `[]byte` elements, create 0 byte slices when a 0 sized binary blob is read.

This makes `[]byte` fully symmetric for roundtrips.

(Tests regression and also that fields are nilled correctly)

Fixes #362

Decoder:
```
		case "abyteslice":
			if msgp.IsNil(bts) {
				bts = bts[1:]
				z.AByteSlice = nil
			} else {
				z.AByteSlice, bts, err = msgp.ReadBytesBytes(bts, z.AByteSlice)
				if err != nil {
					err = msgp.WrapError(err, "AByteSlice")
					return
				}
				if z.AByteSlice == nil {
					z.AByteSlice = make([]byte, 0)
				}
			}
```

Unmarshal:
```
		case "abyteslice":
			if msgp.IsNil(bts) {
				bts = bts[1:]
				z.AByteSlice = nil
			} else {
				z.AByteSlice, bts, err = msgp.ReadBytesBytes(bts, z.AByteSlice)
				if err != nil {
					err = msgp.WrapError(err, "AByteSlice")
					return
				}
				if z.AByteSlice == nil {
					z.AByteSlice = make([]byte, 0)
				}
			}
```

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -278,3 +278,24 @@ type NonMsgStructTags struct {
 }
 
 type EmptyStruct struct{}
+
+type StructByteSlice struct {
+	ABytes      []byte       `msg:",allownil"`
+	AString     []string     `msg:",allownil"`
+	ABool       []bool       `msg:",allownil"`
+	AInt        []int        `msg:",allownil"`
+	AInt8       []int8       `msg:",allownil"`
+	AInt16      []int16      `msg:",allownil"`
+	AInt32      []int32      `msg:",allownil"`
+	AInt64      []int64      `msg:",allownil"`
+	AUint       []uint       `msg:",allownil"`
+	AUint8      []uint8      `msg:",allownil"`
+	AUint16     []uint16     `msg:",allownil"`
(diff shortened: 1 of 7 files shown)


philhofer:
Adopting this would create some odd asymmetry with the default behavior on the encoder side of things which always encodes slices as zero-length arrays, even if they are nil.

IMO if we are going to introduce a behavioral change, we should do a major version bump and clean up all of that stuff at once.

klauspost:
@philhofer Yeah, I see your point.  Maybe we should put it on the `allownil` generation, so it will apply only to those? That would AFAICT make it fully symmetric.

Let me see if that is feasible.

klauspost:
@philhofer Description updated. General change reverted and only `allownil` functionality has changed. PTAL when possible.

philhofer (approved):

PR #364, opened 2024-09-28
Add native json.Number support

### LINKED ISSUE #292: Support json.Number type
I want to covert my json encoded data to msgp data, Then I have to Unmarshal the bytes to a map[string]interface{}, and marshal it to msgp. 
but the data from json Unmarshal do not support json.Number type when I marshal it. 
I have to use json.Number because there are some big int64 numbers in the original bytes.

klauspost:
Allow encoding and decoding of `json.Number` values, either as struct members or as interface members.

Numbers will be encoded as integer, if possible, otherwise float64 is used. The zero value json.Number will be encoded as 0.

It is possible to encode as string with `//msgp:replace json.Number with:string`.

Fixes #292

### Review notes

~I think the only contentious part is that it refuses to encode invalid values, which seems reasonable to me, so the output will either be float or integer or an empty string.~

~Empty string or null, for a "zero value" (not 0)? A bit undecided on this. Open to changes.~

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -1,6 +1,7 @@
 package _generated
 
 import (
+	"encoding/json"
 	"os"
 	"time"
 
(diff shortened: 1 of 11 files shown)


philhofer:
> Empty string or null, for a "zero value" (not 0)? A bit undecided on this. Open to changes.

The behavior of `encoding/json` with `json.Number` is to use zero as the zero value, so that's probably the best option.

klauspost:
Good point. Also makes the output more reliable and guaranteed to be a number type.

klauspost:
Amended, and now will also optionally use float32 as per #365

philhofer (approved):

PR #365, opened 2024-09-29
Use float32 when lossless for float64

klauspost:
When a float64 can be fully represented as a float32, use that instead.

Encoding is backwards compatible with existing reader. `AppendFloat64Only` is made available to unconditionally write a float64.

Micro-benchmarks are not great:

```
BenchmarkAppendFloat64Only-32    	511061935	         2.324 ns/op	3442.55 MB/s	       0 B/op	       0 allocs/op
BenchmarkAppendFloat64-32    	136697270	         8.704 ns/op	 919.08 MB/s	       0 B/op	       0 allocs/op
```

But microbenchmarks are micro-benchmarks. We take a much bigger cost on `AppendInt64` and other integer functions.

We could make it opt-in, but that would make it pretty much unused.

### DIFF
--- a/msgp/edit_test.go
+++ b/msgp/edit_test.go
@@ -120,7 +120,7 @@ func TestReplace(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if !reflect.DeepEqual(m["thing_two"], 4.0) {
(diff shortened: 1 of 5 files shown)


klauspost:
Replaced by #366

PR #374, opened 2024-10-22
Fix allownil propagated to byte slices as slice elements

klauspost:
`[][]byte` would assume that the byte slices had allownil set. This would cause an extra `if x == nil {x = make([]byte,0)}` to be inserted on elements.

Regression from #363


craig65535:
@klauspost Could a new release be made with this fix?

klauspost:
@craig65535 👍🏼 https://github.com/tinylib/msgp/releases/tag/v1.2.3

craig65535:
@klauspost Thank you!

PR #380, opened 2024-11-16
Fix typo, causing extensions to produce invalid json

pietervanheijningen:
Hey!

`msgp.CopyToJSON` can produce invalid json if the msgpack contains an extension. 
Looks to be a typo. Here's the fix

Didn't extensively test, but seems to fix it for the data i throw at it.

Thanks for the project! Just this function to convert to a JSON string has been quite helpful for me


klauspost (approved):
Lgtm

PR #382, opened 2024-12-08
Ignore complexity of children in maps/slices

### LINKED ISSUE #381: omitempty has no effect when node complexity exceeds inlining threshold
Thank you for publishing this lovely library. :)

This bug smells similar to #376.  It is so similar that I did double check I was on trunk before sending this bug report.

```
% dlv exec $(command -v msgp)
Type 'help' for list of commands.
(dlv) sources
[...]
/Users/saj/Documents/src/go/pkg/mod/github.com/tinylib/msgp@v1.2.5-0.20241117134132-901ed0042e54/gen/decode.go
/Users/saj/Documents/src/go/pkg/mod/github.com/tinylib/msgp@v1.2.5-0.20241117134132-901ed0042e54/gen/elem.go
/Users/saj/Documents/src/go/pkg/mod/github.com/tinylib/msgp@v1.2.5-0.20241117134132-901ed0042e54/gen/encode.go
[...]
```

---

The following input works as expected.

```go
package main

//go:generate msgp

type Flibbity struct {
	A Flibs `msg:"a,omitempty"`
	B Flibs `msg:"b,omitempty"`
}

type Flibs []Flib

type Flib struct {
	X int8 `msg:"x"`
	Y int8 `msg:"y"`
}
```

`Flibbity.MarshalMsg` includes the omitempty tests.

```go
	// check for omitted fields
	zb0001Len := uint32(2)
	var zb0001Mask uint8 /* 2 bits */
	_ = zb0001Mask
	if z.A == nil {
		zb0001Len--
		zb0001Mask |= 0x1
	}
	if z.B == nil {
		zb0001Len--
		zb0001Mask |= 0x2
	}
	// variable map header, size zb0001Len
```

---

The following input demonstrates the problem.

```go
package main

//go:generate msgp

type Flobbity struct {
	A Flobs `msg:"a,omitempty"`
	B Flobs `msg:"b,omitempty"`
}

type Flobs []Flob

type Flob struct {
	X Numberwang `msg:"x"`
	Y int8       `msg:"y"`
}

type Numberwang int8
```

`Flobbity.MarshalMsg` is missing the omitempty tests.

```go
	// check for omitted fields
	zb0001Len := uint32(2)
	var zb0001Mask uint8 /* 2 bits */
	_ = zb0001Mask
	// variable map header, size zb0001Len
```

Curiously, if we remove either one of `A` or `B` from the type definition of `Flobbity`, the second example also begins to work as expected.

Here is a repository with the sources above.
https://github.com/saj/tinylib-msgp-381

klauspost:
When considering inlining maps/slices/arrays, apply a fixed cost and ignore the cost of the children.

If children are too expensive, they will not be inlined, but it shouldn't affect whether the map itself is inlined.

This only really applies when a map or slice type is aliased, otherwise it will not be considered for inlining.

Fixes #381

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -308,3 +308,19 @@ type NumberJSONSample struct {
 	Map    map[string]json.Number
 	OE     json.Number `msg:",omitempty"`
 }
+
+type Flobbity struct {
+	A Flobs `msg:"a,omitempty"`
+	B Flobs `msg:"b,omitempty"`
+}
+
+type Flobs []Flob
+
+type Flob struct {
+	X Numberwang `msg:"x"`
+	Y int8       `msg:"y"`
+	Z int8       `msg:"z"`
+	W int32      `msg:"w"`
+}
+
+type Numberwang int8
--- a/gen/elem.go
+++ b/gen/elem.go
@@ -268,7 +268,10 @@ func (a *Array) Copy() Elem {
 	return &b
 }
 
-func (a *Array) Complexity() int { return 1 + a.Els.Complexity() }
+func (a *Array) Complexity() int {
+	// We consider the complexity constant and leave the children to decide on their own.
+	return 2
+}
 
 // ZeroExpr returns the zero/empty expression or empty string if not supported.  Unsupported for this case.
 func (a *Array) ZeroExpr() string { return "" }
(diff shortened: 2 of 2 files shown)


philhofer:
At some point it might be worth checking if the inlining actually does anything for performance at this point.

Back when I wrote this code in 2014 the Go compiler had awful inlining heuristics, and couldn't do "mid-stack inlining" at all. That's been fixed for a few years now.

philhofer (approved):

klauspost:
@philhofer To be honest, in my experience the in-liner is still quite bad. And a proper solution has been pushed down the road for years, with PGO promising magic improvements that are only useful in extremely limited scenarios.

I would *love* to be able to inline the [Next](https://github.com/philhofer/fwd/pull/32/files#diff-c34cdbc81a6de518e5a2d06b326bfc4179ff5e6b996927ea9232fc0340a424f2R291-R300) call manually, so only the `r.next()` would be a function call.

Thanks for taking a look. I am pretty sure this will be fine performance-wise - and probably a net zero or minor gain.

PR #386, opened 2025-02-10
fix: Add handler for EndlessReader possible infinite loop

### LINKED ISSUE #385: State of using zero-length arrays for Reader
If we pass empty slice we will get infinite loop:
https://github.com/tinylib/msgp/blob/3867c342e66c5f1e99020342f4f1b7f073cce5ec/msgp/circular.go#L20

Also we can have zero div error:

https://github.com/tinylib/msgp/blob/3867c342e66c5f1e99020342f4f1b7f073cce5ec/msgp/circular.go#L39

Maybe we should restrict assign empty buffers for Reader?

wooffie:
Closes #385

### DIFF
--- a/msgp/circular.go
+++ b/msgp/circular.go
@@ -14,8 +14,12 @@ type EndlessReader struct {
 	offset int
 }
 
-// NewEndlessReader returns a new endless reader
+// NewEndlessReader returns a new endless reader.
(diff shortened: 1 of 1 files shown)


klauspost (approved):
lgtm

PR #389, opened 2025-03-09
Add zero-copy Unmarshal

klauspost:
Adds `zerocopy` tag that will allow byte slices to be unmarshaled with zero-copy, referencing the original slice.

AFAICT not possible for other types or when decoding.


philhofer:
LGTM.

(Not an issue with this PR per se, but perhaps we should document all of these field tags in the `go doc` output. There has been quite a proliferation of them and some of them have rather subtle behavioral implications.)

PR #392, opened 2025-05-06
Add directive to decode/unmarshal to UTC

klauspost:
Adds `//msgp:timezone utc` that will decode and unmarshal to UTC time.

Default is `//msgp:timezone local`. If someone is feeling adventurous they can expand on these 2 settings, but this should cover 99.99% of use cases.

Avoids nasty hacks, time setting `time.Local` or having to do reflection to go through structs.

Want for a new function, since it was a much cleaner solution than trying to convert after reading.

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -20,6 +20,8 @@ import (
 // a struct definition is
 // by adding it to this file.
 
+//msgp:timezone utc
+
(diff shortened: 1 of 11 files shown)


marktheunissen:
+1 I've tested this and it has worked, thanks @klauspost

Issue #345, opened 2024-04-27
Make `shim` directive handle map key


phqb:
I'm using `tinylib/msgp` to add marshaling/unmarshaling to existing structs. This library works very well in my usecase. I appreciate `tinylib/msgp`'s maintainers a lot. 

My structs have a lot of `map` whose keys are not `string` and MessagePack requires map key to be string. I can replace all these map keys with `string` but doing this makes my structs lost some of their readability and I have to do some refactoring.

For an example, I have to change the following struct:
```
type Foo <some type can convert to/from string> 

type MyStruct struct {
    Map map[Foo]*Bar
}
```
to
```
type MyStruct struct {
    MapByFoo map[string]*Bar
}
```
to maintain readability.

`tinylib/msgp` already has `shim` directive that allows any type to be (un)marshaled as another type. I use this feature a lot to avoid refactoring my structs. **It will be nice that `shim` directive also transform non-string map key to string** so the following code:
```
//msgp:shim Foo as:string using:FooToString/StringToFoo

type MyStruct struct {
    Map map[Foo]*Bar
}
```
will be generated as
```
// EncodeMsg implements msgp.Encodable
func (z *MyStruct) EncodeMsg(en *msgp.Writer) (err error) {
        ...
	for za0001, za0002 := range z.Map {
		err = en.WriteString(FooToString(za0001)) <-----
		if err != nil {
			err = msgp.WrapError(err, "Map")
			return
		}
		if za0002 == nil {
			err = en.WriteNil()
			if err != nil {
				return
			}
		} else {
			...
		}
	}
	return
}

// DecodeMsg implements msgp.Decodable
func (z *MyStruct) DecodeMsg(dc *msgp.Reader) (err error) {
	var field []byte
	_ = field
	var zb0001 uint32
	zb0001, err = dc.ReadMapHeader()
	if err != nil {
		err = msgp.WrapError(err)
		return
	}
	for zb0001 > 0 {
		zb0001--
		field, err = dc.ReadMapKeyPtr()
		if err != nil {
			err = msgp.WrapError(err)
			return
		}
		switch msgp.UnsafeString(field) {
		case "Map":
			var zb0002 uint32
			zb0002, err = dc.ReadMapHeader()
			if err != nil {
				err = msgp.WrapError(err, "Map")
				return
			}
			if z.Map == nil {
				z.Map = make(map[string]*Bar, zb0002)
			} else if len(z.Map) > 0 {
				for key := range z.Map {
[...]

Issue #305, opened 2021-12-17
`map[int]int` failed to parse


fananchong:
My code is :
```go
package main

const Eight = 8

type MyInt int
type Data []byte
type Test1 map[string]int
type Test2 map[int]int

type Struct struct {
	Which  map[int]*MyInt
	Which1 Test1
	Which2 Test2
	Other  Data
	Nums   [Eight]float64
}

func main() {

}
```

and, shell cmd like this:

```shell
fananchong@ubuntu-vm:~/msgp_test$ msgp -file=main.go -o=main_msgp.go -v
Input: "main.go"
warn: main.go: Struct: Which: ignored
warn: main.go: Test2: failed to parse
info: main.go: Struct: inlining Test1
warn: main.go: Struct: unresolved identifier: Test2
info: main.go: Struct: inlining Data
Wrote and formatted "main_msgp.go"
Wrote and formatted "main_msgp_test.go"
fananchong@ubuntu-vm:~/msgp_test$ go build
# msgp_test
./main_msgp.go:162:18: z.Which2.DecodeMsg undefined (type Test2 has no field or method DecodeMsg)
./main_msgp.go:236:16: z.Which2.EncodeMsg undefined (type Test2 has no field or method EncodeMsg)
./main_msgp.go:284:19: z.Which2.MarshalMsg undefined (type Test2 has no field or method MarshalMsg)
./main_msgp.go:350:23: z.Which2.UnmarshalMsg undefined (type Test2 has no field or method UnmarshalMsg)
./main_msgp.go:404:19: z.Which2.Msgsize undefined (type Test2 has no field or method Msgsize)
```

![image](https://user-images.githubusercontent.com/20277955/146512232-8480913b-1764-4a00-bfe8-f54a5e8fdbe0.png)

klauspost:
Added in #398

Issue #257, opened 2019-06-27
maps with int keys


skaldesh:
Hey, I was wondering, why does the code generator ignore my map in this struct?
```go
type Test struct {
    M map[int]int `msgp:"m"`
}
```
As far as I know, the message pack spec allows such keys for maps

philhofer:
You're right that the spec allows for such keys. (Actually, last time
I looked at the spec, it allows even nonsensical objects like arrays
and float64 NaNs to be map keys...)

Presently, the code only implements "string-like" keys, as that
preserves transparent JSON interoperability; there's a caveat to this
effect in the README.

This is something I'd consider fixing for a v2 revision, but probably
not as part of v1.

jiaojianbupt:
Any update?

PR #398, opened 2025-08-10
Add binary map key and shimming

klauspost:
* Allows for binary map keys
* Allows for non-string map keys to be shimmed as well as cast from string type aliases.

Adds directive `maps`

* `//msgp:maps shim|binary`

Not adding the directive will retain current behaviour where maps without string keys are ignored.

Adding 'shim' will require you to add shims to custom map key types. Shims must be with string type. See `_generated/map_shim_key.go` for examples.

Adding `binary` will allow for binary marshaling of map keys. Types are inferred with standard rules. Binary encoding can be shimmed and can return any standard type. See `_generated/map_bin_key.go` for examples.

Updates/Resolves: #331 #345 #305 #257 #232

### DIFF
--- a/_generated/def.go
+++ b/_generated/def.go
@@ -342,3 +342,7 @@ type Flob struct {
 }
 
 type Numberwang int8
+
+//msgp:ignore ExternalString
+type ExternalString string
+type ExternalArr [4]byte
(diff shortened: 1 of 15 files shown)


klauspost:
(considered adding automatic shimming of built-in types as an option, but leaving that for later)

klauspost:
Added. top post updated.

PR #410, opened 2025-09-17
feat: Support generic types

klauspost:
This adds support for generic types.

Say you have `type Foo[T any] struct`, you must add a constraint that it can be handled properly. The constraint is added with a built-in `type Foo[T any, _ msgp.RTFor[T]`.

If T supports serialization this will work. Multiple types can be added, but each must have the constraint.

Nested of types are supported. For example:

```go
type Foo[T any, P msgp.RTFor[T]] struct {
    X  Bar[T, P] `msg:",allownil"`
}

type Bar[T any, _ msgp.RTFor[T]] struct {...}
```

Notice how `P` can be used for a nested generic type.

## Limitations

* Types cannot be primitives since they must support marshalling. It can however be an `type Int64 int` with generation should still work
* Since we cannot reliably create instances with arbitrary types there are no tests.
* The `msgp.RTFor[T]` type shouldn't be used for types in the struct since new instances cannot be created and can cause a runtime crash.
* There will be warnings like: `warn: generics.go: GenericTest: A: possible non-local identifier: T`.
* Some directives may exhibit unexpected behaviour. Not tested too deeply.


klauspost:
Ping @philhofer - Any objections to this being merged?

PR #418, opened 2025-11-13
Add support for aliased map/slice/arrays

### LINKED ISSUE #416: Generics: Instantiation of the list of generic structs is generated w/out generics
Generics support in the new version is great, but I've found an issue with the generated code for a list case: if you add a list type with generics then its instantiation in the generated code is missing generics definition.

The easiest way to reproduce the issue is to add

```go
type GenericTest3List[A, B any, AP msgp.RTFor[A], BP msgp.RTFor[B]] []GenericTest3[A, B, AP, BP]
```

to the https://github.com/tinylib/msgp/blob/master/_generated/generics.go and generate the code - `DecodeMsg` and `UnmarshalMsg` are generated with the following instantiation code:

```go
	if cap((*z)) >= int(zb0002) {
		(*z) = (*z)[:zb0002]
	} else {
		(*z) = make(GenericTest3List, zb0002)
	}
```

while it should be

```go
	if cap((*z)) >= int(zb0002) {
		(*z) = (*z)[:zb0002]
	} else {
		(*z) = make(GenericTest3List[A, B, AP, BP], zb0002)
	}
```

klauspost:
Fixes #416

### DIFF
--- a/gen/encode.go
+++ b/gen/encode.go
@@ -346,7 +346,7 @@ func (e *encodeGen) gBase(b *BaseElem) {
 		}
 
 		// Strip type parameters from dst for lookup in ToPointerMap
-		lookupKey := dst
+		lookupKey := stripTypeParams(dst)
 		if idx := strings.Index(dst, "["); idx != -1 {
 			lookupKey = dst[:idx]
 		}
--- a/gen/size.go
+++ b/gen/size.go
@@ -249,7 +249,7 @@ func (s *sizeGen) gBase(b *BaseElem) {
 		}
 
 		// Strip type parameters from dst for lookup in ToPointerMap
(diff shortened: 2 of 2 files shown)


gen/encode.go
```diff
@@ -344,7 +344,14 @@ func (e *encodeGen) gBase(b *BaseElem) {
 		if b.typeParams.isPtr {
 			dst = "*" + dst
 		}
-		if remap := b.typeParams.ToPointerMap[dst]; remap != "" {
+
+		// Strip type parameters from dst for lookup in ToPointerMap
+		lookupKey := dst
```
philhofer:
```suggestion
		lookupKey := stripTypeParams(dst)
```

PR #423, opened 2025-11-20
Allow ignore directives to contain regex

klauspost:
Prefixing name of a an ignore directive with `regex:` will make it ignore all RE2 regex matches.

Examples:

```
//msgp:ignore regex:IGNORE
```
Ignore all types that contains `IGNORE`.

```
//msgp:size ignore regex:IGNORE
```
Do not generate `Msgsize() int` for all types that contains IGNORE.

### DIFF
--- a/gen/spec.go
+++ b/gen/spec.go
@@ -129,7 +129,7 @@ type TransformPass func(Elem) Elem
 // IgnoreTypename is a pass that just ignores
 // types of a given name.
 func IgnoreTypename(name string) TransformPass {
-	if strings.HasPrefix(name, "regex:") {
+	if name, ok := strings.CutPrefix(name, "regex:"); ok {
 		name = strings.TrimPrefix(name, "regex:")
 		rx, err := regexp.Compile(name)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


gen/spec.go
```diff
@@ -128,6 +129,19 @@ type TransformPass func(Elem) Elem
 // IgnoreTypename is a pass that just ignores
 // types of a given name.
 func IgnoreTypename(name string) TransformPass {
+	if strings.HasPrefix(name, "regex:") {
```
philhofer:
```suggestion
	if name, ok := strings.CutPrefix(name, "regex:"); ok {
```

parse/directives.go
```diff
@@ -167,6 +168,22 @@ func ignore(text []string, f *FileSet) error {
 	}
 	for _, item := range text[1:] {
 		name := strings.TrimSpace(item)
+		if strings.HasPrefix(name, "regex:") {
```
philhofer:
```suggestion
		if name, ok := strings.CutPrefix(name, "regex:"); ok {
```

gen/spec.go
klauspost: Ah yeah. Knew there was one for it, but managed to overlook it :)

PR #428, opened 2025-12-04
Update fuzz tests plus minor fixes

klauspost:
Decided to make ReadArray always stop after an error. There is a too high risk of running into an infinite loop.

Added size checks for (Writer.)WriteBytes/WriteString/WriteStringFromBytes

Add panic recovery to binary marshal helpers.

### DIFF
--- a/msgp/write.go
+++ b/msgp/write.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"math"
 	"reflect"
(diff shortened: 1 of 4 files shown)


msgp/write.go
```diff
@@ -447,6 +448,9 @@ func (mw *Writer) WriteUint(u uint) error { return mw.WriteUint64(uint64(u)) }
 
 // WriteBytes writes binary as 'bin' to the writer
 func (mw *Writer) WriteBytes(b []byte) error {
+	if uint64(len(b)) > math.MaxInt32 {
```
philhofer: Shouldn't this be `math.MaxUint32`?

klauspost: You are absolutely right! :facepalm:

Issue #425, opened 2025-11-21
Cleanup: Add "msgp.ReadByteLimit" to simplify codegen for bin field decoding


klauspost:
As observed in https://github.com/tinylib/msgp/pull/424 we should probably add a function that allows for reading bytes.

```
func (r *Reader) ReadBytesLimit(dst []byte, limit int64) ([]byte,error)
```

... is probably something like what we want.

PR #432, opened 2025-12-22
Add (Reader).ReadBytesLimit

klauspost:
Simplifies generated code.

Fixes https://github.com/tinylib/msgp/issues/425

Modify a few tests to allow wrapped errors.


klauspost:
Ah. I see you sent in the fix. Made another release

PR #435, opened 2026-03-16
msgp: Reader: reset limits on cached readers

philhofer:
A hand-written implementation of DecodeMsg may adjust the reader limits inside msgp.CopyToJSON, msgp.Decode, and so forth. Ensure that a Reader pulled from the pool has the same limits (zeros) every time.


klauspost (approved):
👍🏼

PR #437, opened 2026-04-01
Add support for aliases on tags

natewernimont-wk:
I maintain usages of msgp that initially neglected to specify short name tags. As such, their full msgp tag is a quite long name. In order to optimize the byte allocations of these objects, I'd like to move the serialized tag to a new name. Doing so today would require manually editing the generated code or hand-rolling our own.

This PR adds the ability to add an `alias=<oldName1>;<oldName2>;...` tag to msgp labels. Aliases are only supported on deserialization - they do not affect the key at serialization time.

### DIFF
--- a/_generated/alias.go
+++ b/_generated/alias.go
@@ -1,0 +1,21 @@
+package _generated
+
+//go:generate msgp
+
+type AliasedFieldsV3 struct {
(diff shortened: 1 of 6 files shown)


natewernimont-wk:
rejected in issue

PR #441, opened 2026-05-04
msgp: Update Buffered method to return write location instead of remaining buffer space

### LINKED ISSUE #440: `Writer.Buffered()` returns free space instead of bytes currently buffered
`msgp.Writer.Buffered()` is documented as returning the number of bytes in the write buffer, but the implementation returns `len(mw.buf) - mw.wloc`, which is remaining capacity/free space.

This is inconsistent with Go’s `bufio.Writer.Buffered()`, which returns the number of bytes currently buffered.

Expected: `return mw.wloc`
Actual: `return len(mw.buf) - mw.wloc`

Tijani-Dia:
Fixes #440

### DIFF
--- a/msgp/write.go
+++ b/msgp/write.go
@@ -183,7 +183,7 @@ func (mw *Writer) flush() error {
 func (mw *Writer) Flush() error { return mw.flush() }
 
 // Buffered returns the number bytes in the write buffer
-func (mw *Writer) Buffered() int { return len(mw.buf) - mw.wloc }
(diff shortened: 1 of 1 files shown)


klauspost (changes requested):
No. We can't change it. As I wrote you are welcome to update the docs.

Tijani-Dia:
I will close it then as I don't feel right saying in docs that `Buffered` returns available space - due to the method name itself and how it works in `bufio`.

PR #443, opened 2026-05-11
Add JSON Marshal/Unmarshal to setof

klauspost:
Each type not also support JSON serializers so sets are stored as JSON. Sorted/Unsorted also applies here.

Mildly breaking. Open for alternatives.

### DIFF
--- a/msgp/setof/json.go
+++ b/msgp/setof/json.go
@@ -22,6 +22,7 @@ func skipWS(data []byte, i int) int {
 
 // jsonArrayIter calls fn for each raw element in a JSON array.
 // String elements include their surrounding quotes. Does not allocate.
+// Simple string and integer types are supported.
 func jsonArrayIter(data []byte, fn func(raw []byte) error) error {
 	i := skipWS(data, 0)
 	if i >= len(data) || data[i] != '[' {
--- a/msgp/setof/json.go
+++ b/msgp/setof/json.go
@@ -80,6 +80,7 @@ func jsonAppendQuote(dst []byte, s string) []byte {
 	dst = append(dst, '"')
 	for _, r := range s {
 		if r >= 0x20 && r != '"' && r != '\\' {
+			// Any printable rune...
 			dst = utf8.AppendRune(dst, r)
 			continue
 		}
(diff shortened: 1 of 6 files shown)


msgp/setof/json.go
```diff
@@ -0,0 +1,152 @@
...
+					continue
+				}
+				if data[i] == '"' {
+					i++
+					break
+				}
+				i++
+			}
+		} else {
+			for i < len(data) && data[i] != ',' && data[i] != ']' &&
```
philhofer, line 55: I think you only ever use this function to parse numbers and strings, so maybe it's not a huge deal, but obviously this fails in rather pernicious ways if you have a sub-array or sub-structure here.

philhofer: (I really wish the Go team would officially release the encoding/json/v2 API because it has a public API for the lexer we could use... oh well.)

klauspost: I think it is scheduled for v1.27, but haven't followed too closely.

But yeah, made it specifically for the cases we cover.

klauspost: (added comment to function for future caution)

msgp/setof/json.go
```diff
@@ -0,0 +1,153 @@
...
+		case '\f':
+			dst = append(dst, '\\', 'f')
+		case '\n':
+			dst = append(dst, '\\', 'n')
+		case '\r':
+			dst = append(dst, '\\', 'r')
+		case '\t':
+			dst = append(dst, '\\', 't')
+		default:
+			dst = append(dst, '\\', 'u', '0', '0', hexDigit(byte(r>>4)), hexDigit(byte(r&0xf)))
```
philhofer, line 102: Doesn't JSON support the full 4 digits here? (Or, conversely, if it doesnt, then presumably you'd want to emit U+FFFD here?)

klauspost: You may very well be right. Actually, let me set up a fuzz test. Way easier to find  roundtrip problems/disceptencies.

klauspost: This should only be reachable for control runes, so < 0x20. I've added fuzz tests to see if it is roundtrip equivalent to `encoding/json`. Added clarification.

philhofer (approved):

klauspost:
Merged, but holding off release for a bit more.

PR #444, opened 2026-06-03
Allow prioritized list in tag directive

klauspost:
Tags are tried in order and we maintain the existing fallback.

Adds `tags` as alias.

### DIFF
--- a/parse/directives.go
+++ b/parse/directives.go
@@ -48,6 +48,7 @@ var directives = map[string]directive{
 // and then add it to this list.
 var earlyDirectives = map[string]directive{
 	"tag":     tag,
+	"tags":    tag,
 	"pointer": pointer,
 	"maps":    maps,
 }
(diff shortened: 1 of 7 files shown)


parse/directives.go
```diff
@@ -237,12 +238,28 @@ func asvartuple(text []string, f *FileSet) error {
...
+			}
+		}
+	}
+	if len(names) == 0 {
 		return nil
 	}
-	f.tagName = strings.TrimSpace(text[1])
-	infof("using field tag %q\n", f.tagName)
+	f.tagNames = names
+	infof("using field tag %q\n", names)
```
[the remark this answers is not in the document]
klauspost: ack

parse/directives.go
```diff
@@ -237,12 +238,28 @@ func asvartuple(text []string, f *FileSet) error {
...
+// read from the first tag that has a non-empty value, falling back to msg
+// then msgpack if none of the listed tags match.
 func tag(text []string, f *FileSet) error {
-	if len(text) != 2 {
+	if len(text) < 2 {
+		return nil
+	}
+	var names []string
+	for _, t := range text[1:] {
+		for _, n := range strings.Split(t, ",") {
```
philhofer: `for n := range strings.SplitSeq(t, ",")`

parse/getast.go
```diff
@@ -527,15 +527,18 @@ func (fs *FileSet) getField(f *ast.Field) []gen.StructField {
...
+		st := reflect.StructTag(strings.Trim(f.Tag.Value, "`"))
 		var body string
-		if fs.tagName != "" {
-			body = reflect.StructTag(strings.Trim(f.Tag.Value, "`")).Get(fs.tagName)
+		for _, name := range fs.tagNames {
+			if body = st.Get(name); body != "" {
+				break
+			}
 		}
 		if body == "" {
```
philhofer: This code would be a little simpler if `tagNames` always ended in `[..., "msg", "msgpack"]`. Then the behavior of the `tags` directive is to just prepend entries to the list.

parse/directives.go
klauspost: Oh yeah, we are in 1.24 land now :)

(not going to admit I didn't know about it)

parse/getast.go
klauspost: Good suggestion!

klauspost: I kept the "one directive per file, last directive wins" behaviour as before. 

It would be a small breakage, but I don't see much reason for it.

philhofer (approved):

PR #445, opened 2026-07-23
fix: NewWriter(nil) uses io.Discard

sonnemusk:
## What

`NewWriter(nil)` uses `io.Discard` instead of keeping a nil writer.

## Why

Optional encode sinks should not panic on Flush/Write.

## Test

- `TestNewWriterNil`

### DIFF
--- a/msgp/nil_writer_test.go
+++ b/msgp/nil_writer_test.go
@@ -1,0 +1,13 @@
+package msgp
+
+import "testing"
+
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing this PR. After review, this is too low-value / looks like bulk automated noise rather than a carefully chosen change. Sorry for the noise — will focus on real, high-signal fixes only.

PR #450, opened 2026-08-14
Fix ReadMapStrIntf rejecting bin-typed map keys

### LINKED ISSUE #449: ReadMapStrIntf does not accept 'bin' typed map keys, unlike the rest of the library
[Reader.ReadMapStrIntf](https://github.com/tinylib/msgp/blob/801fd461b530b0eb2d527c4123a5c9681442e62e/msgp/read.go#L1367) rejects a msgpack map whose key is encoded as bin instead of str, even though the library has documented and implemented support for this case since 2014.  The behavior is implemented in [Reader.ReadMapKey](https://github.com/tinylib/msgp/blob/801fd461b530b0eb2d527c4123a5c9681442e62e/msgp/read.go#L467). [ReadMapStrIntf](https://github.com/tinylib/msgp/blob/801fd461b530b0eb2d527c4123a5c9681442e62e/msgp/read.go#L1383) doesn't call it.  A TypeError is returned for any map with a bin typed key, and ReadMapKey sits unreferenced.

Context:

- ReadMapStrIntf was added on 2014-11-02 (3bda8c7), with the bare ReadString() call.
- Issue #27 was opened the next day (2014-11-03), asking for bin-encoded map keys to be tolerated.
- It was fixed same-day: 18a67e9 ("accept 'bin' as a key in encoded values") added ReadMapKey, and 0d8d272 the next morning extended it to JSON translation and documented it in the README as the intentional, supported behavior described above.
- ReadMapStrIntf, written the day before that fix, was apparently never updated to use it. Issue #27 itself was never closed/linked to either commit.

The proposed fix is to swap the bare ReadString() call for the existing ReadMapKey helper:

```
for i := uint32(0); i < sz; i++ {
	var keyBytes []byte
	var val any
	keyBytes, err = m.ReadMapKey(nil)
	if err != nil {
		return
	}
	val, err = m.ReadIntf()
	if err != nil {
		return
	}
	mp[string(keyBytes)] = val
}
```

Found this via open-telemetry/opentelemetry-collector-contrib's [fluentforwardreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/50235), which decodes nested map values via ReadIntf -> ReadMapStrIntf. Its own top-level record decode already has a hand-rolled bin-fallback for exactly this reason , but anything nested goes through ReadMapStrIntf and hits this gap where a single bin-tagged key anywhere in a nested map fails the entire decode, taking the whole batch (PackedForward mode) down with it, not just that one field. A sender that tags strings ASCII-8BIT before serializing (e.g. Fluentd's own `-Eascii-8bit:ascii-8bit ` default when its JSON parser is involved) will trip this on any nested field, silently and without a top-level fallback to save it.

Happy to open a PR for this if that's welcome. I believe it's a small, self-contained change.

mleklund:
ReadMapStrIntf reads map keys with a plain ReadString call, which only accepts the 'str' wire type. The rest of the library already accepts 'bin' as well - ReadMapKey and ReadMapKeyZC have supported it since 2014, and this function's own zero-copy sibling, ReadMapStrIntfBytes, already goes through ReadMapKeyZC.

Swaps the ReadString call for ReadMapKey, matching what ReadMapStrIntfBytes already does.

Fixes #449

### DIFF
--- a/msgp/read.go
+++ b/msgp/read.go
@@ -1378,9 +1378,9 @@ func (m *Reader) ReadMapStrIntf(mp map[string]any) (err error) {
 		return
 	}
 	for i := uint32(0); i < sz; i++ {
-		var key string
+		var key []byte
 		var val any
-		key, err = m.ReadString()
+		key, err = m.ReadMapKey(nil)
 		if err != nil {
 			return
 		}
@@ -1388,7 +1388,7 @@ func (m *Reader) ReadMapStrIntf(mp map[string]any) (err error) {
 		if err != nil {
 			return
 		}
-		mp[key] = val
+		mp[string(key)] = val
 	}
 	return
 }
--- a/msgp/read.go
+++ b/msgp/read.go
@@ -1377,10 +1377,10 @@ func (m *Reader) ReadMapStrIntf(mp map[string]any) (err error) {
 		err = ErrLimitExceeded
 		return
 	}
+	var key []byte
 	for i := uint32(0); i < sz; i++ {
-		var key []byte
 		var val any
-		key, err = m.ReadMapKey(nil)
+		key, err = m.ReadMapKey(key[:0])
 		if err != nil {
 			return
 		}
--- a/msgp/read_test.go
+++ b/msgp/read_test.go
@@ -80,6 +80,38 @@ func TestReadIntf(t *testing.T) {
 	}
 }
 
+// both the 'str' and 'bin' types are acceptable map keys, same as TestReadMapKey
+func TestReadMapStrIntfBinKey(t *testing.T) {
+	var buf bytes.Buffer
+	en := NewWriter(&buf)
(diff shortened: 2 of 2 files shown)


klauspost (approved):
lgtm

msgp/read.go
```diff
@@ -1378,17 +1378,17 @@ func (m *Reader) ReadMapStrIntf(mp map[string]any) (err error) {
 		return
 	}
 	for i := uint32(0); i < sz; i++ {
-		var key string
+		var key []byte
```
philhofer: Hoist the declaration of `key` outside of the loop and pass `key[:0]` to `ReadMapKey`. That ought to reduce the number of allocations in this loop from roughly 2*keys to 1*keys.

mleklund: fixed and added a test.

PR #452, opened 2026-09-19
size: fix empty struct collection expressions

### LINKED ISSUE #101: msgp generated code fails to compile: slice of empty struct
ae2d27e6b10e11840e783fd49430fb13d7113ff2 on OSX 10.10.2 go1.5beta1/amd64

Thanks for the #99 fix!  Just one other thing I ran across --

```
jason@jasonsmac ~/tmp $ go generate
======== MessagePack Code Generator =======
>>> Input: "ex.go"...                                                                                              
parsing Stew...                                                                                                    
parsing Empty...                                                                                                   
 ⚠ unable to parse Empty                                                                                           
 ⚠ WARNING: unresolved identifier: Empty                                                                           
>>> Wrote and formatted "ex_gen.go"                                                                                
>>> Wrote and formatted "ex_gen_test.go"                                                                           
>>> Done.                                                                                                          
jason@jasonsmac ~/tmp $ go build
# _/Users/jason/tmp
./ex_gen.go:39: z.Eslice[xvk].DecodeMsg undefined (type Empty has no field or method DecodeMsg)
./ex_gen.go:67: z.Eslice[xvk].EncodeMsg undefined (type Empty has no field or method EncodeMsg)
./ex_gen.go:83: z.Eslice[xvk].MarshalMsg undefined (type Empty has no field or method MarshalMsg)
./ex_gen.go:119: z.Eslice[xvk].UnmarshalMsg undefined (type Empty has no field or method UnmarshalMsg)
./ex_gen.go:138: z.Eslice[xvk].Msgsize undefined (type Empty has no field or method Msgsize)
jason@jasonsmac ~/tmp $
```

the input program, ex.go, contains Empty, which is a struct with no fields. The problem does not occur if I put a dummy field inside Empty, but I would rather be able to leave Empty empty for now, as I'm developing, until it needs a field.

```
package main

//go:generate msgp                                                                                                 

type Empty struct{}

type Stew struct {
    Eslice []Empty
}

func main() {

}
```

jakezwang:
Collections of empty structs using the default map encoding currently produce invalid `Msgsize` expressions such as `len(z.Named) * (1 + )`, causing the generator to fail formatting otherwise valid Go input. Return the encoded header size alone when a struct has no field-size terms.

The regression generates and executes code for named and anonymous empty-struct slices and fixed, zero-length, and nested arrays. It checks the size bound and byte/stream round trips.

Fixes #101.

Validation:
- All native packages, including `_generated`, pass on Go 1.24.13 and pass with `-race` on Go 1.26.6.
- The new generator regression passes with `-race` on Go 1.24.13, 1.25.13, and 1.26.6.
- `make prepare`, formatting, and golangci-lint 2.9.0 pass.

Local validation ran on macOS arm64. TinyGo was unavailable, so `tinygotest` was not run. The available lint version was 2.9.0 rather than CI's pinned 2.5.0.


klauspost (approved):
lgtm

